PR #13, opened 2025-03-31
bump ipv4 routing table size

rcgoodfellow:


dpd/p4/constants.p4
```diff
@@ -5,7 +5,7 @@ const bit<16> L2_ISOLATED_FLAG = 0x8000;
 //is going to need to come either through ATCAM/ALPM or code restructuring.
 const int IPV4_NAT_TABLE_SIZE		= 1024; // nat routing table
 const int IPV6_NAT_TABLE_SIZE		= 1024; // nat routing table
-const int IPV4_LPM_SIZE			= 4096; // ipv4 forwarding table
```
Nieuwejaar, line 8: This should get you a larger table size in the p4 code, but the table test will fail unless you update its constant as well.

Nieuwejaar (approved):

PR #14, opened 2025-04-01
wip

zeeshanlakhani:


zeeshanlakhani:
I'll add some notes to the PR tomorrow morning (to explain through pieces).

dpd-client/tests/integration_tests/common.rs
```diff
@@ -780,6 +793,39 @@ impl Switch {
...
+
+    /// If no client_name is given, this returns the number of packets dropped
+    /// for a given reason.
+    ///
+    /// If a client_name is given, we look up the counter for that client
+    /// name and the given counter name.
+    ///
+    /// If that counter isn't in the set returned by dpd, we return
+    /// an error to the caller.
+    pub async fn get_counter(
```
zeeshanlakhani, line 805: moved these functions in for use beyond the counters test (and having mcast tests all live in that module)

dpd/src/counters.rs
```diff
@@ -88,26 +89,46 @@ const COUNTERS: [CounterDescription; 6] = [
...
     },
+    CounterDescription {
+        id: CounterId::Egress,
+        client_name: "Ingress_Egress",
+        p4_name: "pipe.Ingress.egress_ctr",
+    },
     CounterDescription {
         id: CounterId::DropPort,
-        client_name: "Drop_Port",
+        client_name: "Ingress_Drop_Port",
```
zeeshanlakhani, line 104: *Note*: client name changes to decipher.

dpd/p4/constants.p4
```diff
@@ -58,4 +86,8 @@ const bit<8> DROP_IPV6_TTL_EXCEEDED		= 0x0E;
 const bit<8> DROP_IPV4_UNROUTEABLE		= 0x0F;
 const bit<8> DROP_IPV6_UNROUTEABLE		= 0x10;
 const bit<8> DROP_NAT_INGRESS_MISS		= 0x11;
-const bit<32> DROP_REASON_MAX			= 0x12;
+const bit<8> DROP_MULTICAST_NO_GROUP            = 0x12;
+const bit<8> DROP_MULTICAST_INVALID_MAC         = 0x13;
+const bit<8> DROP_MULTICAST_CPU_COPY            = 0x14;
+const bit<8> DROP_MULTICAST_SOURCE_FILTERED     = 0x15;
+const bit<32> DROP_REASON_MAX			= 0x16;
```
zeeshanlakhani, line 93: *Note*: I did move this around to make MAX post the new drop additions.

dpd/src/api_server.rs
```diff
@@ -3061,6 +3070,200 @@ async fn counter_get(
         .map_err(HttpError::from)
 }
 
+#[derive(Deserialize, Serialize, JsonSchema)]
+pub struct MulticastGroupIdParam {
+    pub group_id: String,
+}
+
+// Implement conversion to mcast::Identifier
+impl MulticastGroupIdParam {
```
rcgoodfellow, line 3079: This feels like it wants to be `impl TryFrom<MulticastGroupIdParam> for mcast::Identifier`

dpd/src/api_server.rs
```diff
@@ -3061,6 +3070,200 @@ async fn counter_get(
...
+    pub fn to_identifier(&self) -> Result<mcast::Identifier, HttpError> {
+        // First try to parse as IP address
```
rcgoodfellow, line 3081: Why not just model `MulticastGroupIdParam` as an enum in the first place?

dpd/src/api_server.rs
```diff
@@ -3061,6 +3070,200 @@ async fn counter_get(
...
+) -> Result<HttpResponseDeleted, HttpError> {
+    let switch: &Switch = rqctx.context();
+
+    mcast::reset(switch)
+        .map(|_| HttpResponseDeleted())
+        .map_err(HttpError::from)
+}
+
+/**
+ * Delete a multicast group configuration by IP address.
```
rcgoodfellow, line 3139: s/IP address/group id/

which can be either an ip address or u16?

dpd/src/api_server.rs
```diff
@@ -3061,6 +3070,200 @@ async fn counter_get(
...
+/**
+ * List all multicast groups.
+ */
+#[endpoint {
+    method = GET,
+    path = "/multicast/groups",
+}]
+async fn multicast_groups_list(
+    rqctx: RequestContext<Arc<Switch>>,
+) -> Result<HttpResponseOk<Vec<mcast::MulticastGroupResponse>>, HttpError> {
```
rcgoodfellow, line 3206: Consider making this a paginated result, as the number of groups may be large.

dpd/src/api_server.rs
```diff
@@ -3061,6 +3070,200 @@ async fn counter_get(
...
+ * List all multicast groups with a given tag.
+ */
+#[endpoint {
+    method = GET,
+    path = "/multicast/tags/{tag}",
+}]
+async fn multicast_groups_list_by_tag(
+    rqctx: RequestContext<Arc<Switch>>,
+    path: Path<TagPath>,
+) -> Result<HttpResponseOk<Vec<mcast::MulticastGroupResponse>>, HttpError> {
```
rcgoodfellow, line 3223: Consider pagination.

dpd/p4/headers.p4
```diff
@@ -226,6 +225,40 @@ struct sidecar_ingress_meta_t {
 	// Used for responding to pings
 	mac_addr_t orig_src_mac;	// source mac address before rewriting
 	ipv4_addr_t orig_src_ipv4;	// original ipv4 source
-
 	ipv4_addr_t orig_dst_ipv4;	// original ipv4 target
 }
+
+struct sidecar_egress_meta_t {
```
rcgoodfellow, line 231: I know that there was metadata here before, but we should probably make an explicit `metadata.p4` and have `headers.p4` just contain actual packet headers.

dpd/p4/sidecar.p4
```diff
@@ -10,14 +10,86 @@ const bit<9> USER_SPACE_SERVICE_PORT = 192;
...
 #include <parser.p4>
 
+// Top-level function to validate IPv4 multicast MAC addresses.
+//
+// Validate the IPv4 multicast MAC address format (RFC 1112, RFC 7042).
+// IPv4 multicast addresses (224.0.0.0/4 or 0xE0000000/4) must
+// use MAC addresses that follow the IANA-assigned OUI format
+// 01:00:5e + 0 as first bit of the second byte,
+// followed by 23 bits derived from the IPv4 address.
+bool validate_ipv4_mcast_mac(
```
rcgoodfellow, line 20: I think we should consider doing this type of validation in parsing. e.g. if we see a multicast mac when parsing the ethernet header, then we require either seeing seeing an L3 protocol that is tied to multicast macs (like LLDP for example) or a multicast ip address subsequently.

Same goes for v6 validation below.

dpd/p4/sidecar.p4
```diff
@@ -56,34 +138,60 @@ control Filter(
...
+				// Interface-local multicast
+				drop_mcast_local_interface();
+				drop_mcast_ctr.count(ig_intr_md.ingress_port);
+				drop_reason_ctr.count(meta.drop_reason);
+			} else if (hdr.ipv6.dst_addr[127:120] == 8w0xff) {
+				bool is_valid = validate_ipv6_mcast_mac(hdr.ethernet.dst_mac, hdr.ipv6.dst_addr);
+				if (is_valid) {
+					if (mcast_scope_v6 == 16w0xff02) {
+						// IPv6 Link-local multicast
+						meta.is_link_local_mcast = true;
```
rcgoodfellow, line 179: Is the reason behind all these derived values in metadata to avoid issues with PHV allocation when using actual packet values as match keys?

dpd/p4/sidecar.p4
```diff
@@ -56,34 +138,60 @@ control Filter(
...
-				meta.drop_reason = DROP_MULTICAST_TO_LOCAL_INTERFACE;
-			} else {
-				meta.multicast = (multicast_scope == 16w0xff02);
-				switch_ipv6_addr.apply();
+			bit<16> mcast_scope_v6 = (bit<16>)hdr.ipv6.dst_addr[127:112];
+			if (mcast_scope_v6 == 16w0xff01) {
+				// Interface-local multicast
+				drop_mcast_local_interface();
+				drop_mcast_ctr.count(ig_intr_md.ingress_port);
+				drop_reason_ctr.count(meta.drop_reason);
```
rcgoodfellow, line 173: I think we should bail at this point and not fall through to `switch_ipv6_addr.apply()`?

dpd/p4/sidecar.p4
```diff
@@ -210,24 +320,25 @@ control Services(
...
+			( 0, false, false, _, true, _, _, false, true, true, false, ICMP_ECHO, _ ) : drop_bad_ping;
+			( 0, false, false, _, true, _, _, false, true, false, true, ICMP6_ECHOREPLY, 0 ) : forward_to_userspace;
+			( 0, false, false, _, true, _, _, false, true, false, true, ICMP6_ECHOREPLY, _ ) : drop_bad_ping;
+			( 0, false, false, _, true, _, _, false, true, false, true, ICMP6_ECHO, 0 ) : ping6_reply;
+			( 0, false, false, _, true, _, _, false, true, false, true, ICMP6_ECHO, _ ) : drop_bad_ping;
+			( 0, false, false, _, _, USER_SPACE_SERVICE_PORT, true, _, _, _, _, _, _ ) : forward_from_userspace;
+			( 0, false, true, true, _, USER_SPACE_SERVICE_PORT, true, _, _, _, _, _, _ ) : forward_from_userspace;
+			( 0, false, false, _, _, _, false, true, _, _, _, _, _ ) : forward_to_userspace;
+			( 0, false, false, _, true, _, _, _, _, _, _, _, _ ) : forward_to_userspace;
+			// Link-local multicast
```
rcgoodfellow, line 340: Since we're changing quite a bit around link-local multicast, we need to test this e2e with ddm.

dpd/p4/sidecar.p4
```diff
@@ -373,66 +490,68 @@ control NatIngress (
...
-	// parameter I've manually computed, or I can create 11 actions, each
-	// of which will subtract a hard-coded constant.  Either seems stupid,
-	// but here we are.
-	// XXX: find a less stupid solution
-	action invert() {
-		meta.l4_length = ~meta.l4_length;
+	/* Invert the source and destination addresses.
+
+	*
+	 * This is a separate file for reusability.
```
rcgoodfellow, line 499: Can this be restructured to have a control that is instantiated in multiple places instead of using the preprocessor to kersplat a few actions and a table definition in place?

dpd/p4/sidecar.p4
```diff
@@ -445,13 +564,26 @@ control NatIngress (
 				meta.l4_dst_port = hdr.icmp.data[31:16];
 			}
 		}
+
 		if (hdr.ipv4.isValid()) {
-			ingress_ipv4.apply();
+			if (meta.is_mcast && !meta.is_link_local_mcast)  {
```
rcgoodfellow, line 569: I think the bifurcations in this apply block are getting varied enough where we should consider having this be driven by a static table instead of all this if/else logic. Similar to the `Services` control.

rcgoodfellow (commented):
Thanks Zeeshan. I've provided some comments on the API and am about half way through the P4 code. Have to run, but wanted to get these initial comments in.

dpd/src/api_server.rs
zeeshanlakhani: yep.

FelixMcFelix (commented):
Thanks for this Zeeshan, some notes so far. I haven't yet made it through `dpd/src/mcast.rs` and the table management code yet.

dpd/src/api_server.rs
zeeshanlakhani: yeah. the api_server has some stuff defined only in API vs jsonschema/serd/de on the structures themselves, but I starting going in that direction, so I agree here.

dpd/src/api_server.rs
zeeshanlakhani: Yeah. Good point. I had this on my TODO, so will do (and it's easier now that I don't track routes separately anymore)

dpd/p4/sidecar.p4
zeeshanlakhani: Good point.

dpd/p4/sidecar.p4
zeeshanlakhani: Yep. Going back to the ingress model, I could make these meta again, but if we add more meta fields, we'll run up against it, which I did.

dpd/p4/sidecar.p4
zeeshanlakhani: yep, i had something like that earlier with `const entries`, will do.

dpd-client/tests/integration_tests/table_tests.rs
```diff
@@ -68,18 +73,40 @@ fn gen_ipv6_cidr(idx: usize) -> Ipv6Net {
     Ipv6Net::new(gen_ipv6_addr(idx), 128).unwrap()
 }
 
+/// Generates valid IPv4 multicast addresses that avoid reserved ranges
+fn gen_ipv4_multicast_addr(idx: usize) -> Ipv4Addr {
+    // Start with 224.1.0.0 to avoid the reserved 224.0.0.0/24 range
```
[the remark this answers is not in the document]
zeeshanlakhani: Good point, should double-check that on validation.

dpd/src/api_server.rs
```diff
@@ -2539,6 +2540,10 @@ async fn reset_all(
         error!(switch.log, "failed to reset route data: {:?}", e);
         err = Some(e);
     }
+    if let Err(e) = route::reset(switch).await {
```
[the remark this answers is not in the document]
zeeshanlakhani: Will clean this up.

dpd/p4/sidecar.p4
```diff
@@ -481,6 +613,14 @@ control NatIngress (
...
 				hdr.vlan.setInvalid();
 			}
+
+			// Multicast packts should be decapped without
+			// a service routing check.
+			if (meta.is_mcast && !meta.is_link_local_mcast
+				&& !meta.nat_egress_hit && hdr.geneve.isValid()
+				&& hdr.geneve.vni != 0) {
+				meta.nat_egress_hit = true;
+			}
```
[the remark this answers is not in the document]
zeeshanlakhani: This is mainly about setting the meta for later. Non mcast traffic goes through the `Services` control, which sets this:

```p4
if (!meta.is_switch_address && meta.nat_ingress_port && !meta.nat_ingress_hit) {
			// For packets that were not marked for NAT ingress, but which
			// arrived on an uplink port that only allows in traffic that
			// is meant to be NAT encapsulated.
			meta.drop_reason = DROP_NAT_INGRESS_MISS;
			ig_dprsr_md.drop_ctl = 1;
		}
		else if (meta.is_switch_address && hdr.geneve.isValid() && hdr.geneve.vni != 0) {
			meta.nat_egress_hit = true;
		}
		else {
			service.apply();
		}
```

which is the meta flag that applies the egress table: https://github.com/oxidecomputer/dendrite/blob/2dcb98e38887b9e9dc43f84832cae5e96c7161f4/dpd/p4/sidecar.p4#L758.

dpd/src/api_server.rs
```diff
@@ -3061,6 +3070,200 @@ async fn counter_get(
...
+            Ok(id) => Ok(mcast::Identifier::GroupId(id)),
+            Err(_) => Err(HttpError::for_bad_request(
+                None,
+                format!(
+                    "'{}' is neither a valid IP address nor a valid group ID",
+                    self.group_id
+                ),
+            )),
+        }
+    }
```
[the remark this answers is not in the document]
zeeshanlakhani: Yeah, I had this as a separate endpoint before b/c I wanted the type. Sadly, dropshot doesn't make this easy to handle generally. I guess I'll split the endpoint.

dpd/src/api_server.rs
```diff
@@ -2551,6 +2556,10 @@ async fn reset_all(
         error!(switch.log, "failed to reset ipv6 nat table: {:?}", e);
         err = Some(e);
     }
+    if let Err(e) = mcast::reset(switch) {
```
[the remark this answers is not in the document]
zeeshanlakhani: This is just the one we need (not the extra route one).

dpd/src/api_server.rs
zeeshanlakhani: I'm re-arranging this, but now I remember why: dropshot doesn't allow it on the param side.

dpd/p4/sidecar.p4
zeeshanlakhani: `apply cannot be called from actions` for triggering a table apply (in an action) from const table, but I can make some of this nicer.

zeeshanlakhani:
Got this back down to 14 stages.

dpd/src/mcast.rs
```diff
@@ -0,0 +1,1271 @@
...
+
+/// Modify a multicast group configuration.
+pub(crate) fn modify_group(
+    s: &Switch,
+    ident: Identifier,
+    new_group_info: MulticastGroupUpdateEntry,
+) -> DpdResult<MulticastGroupResponse> {
+    let (group_ip, group) = find_group(s, &ident)?;
+
+    debug!(s.log, "modifying multicast group for IP {}", group_ip);
```
FelixMcFelix, line 499: We're touching a few tables in a row here, using deltas computed at the start. Is there a risk that concurrent calls to this endpoint from, e.g., our 3 Nexus instances in the rack might try to insert/remove entries in an inconsistent way?

dpd-client/tests/integration_tests/mcast.rs
```diff
@@ -0,0 +1,2381 @@
...
+
+        test_pkts.push(TestPacket {
+            packet: Arc::new(to_send),
+            port: ingress,
+        });
+
+        expected_pkts.push(TestPacket {
+            packet: Arc::new(to_recv.clone()),
+            port: egress1,
+        });
```
FelixMcFelix, line 2040: I think this ties in with my confusion over why we're always following NAT ingress with NAT egress in the multicast case, given that the output packets are only different in their VLAN. I thought a bare frame would come out Geneve encapsulated onto the underlay network (`integration_tests::nat::gen_geneve_packet`)? The behaviour I had maybe expected would be:

* Multicast packet arrives on a front panel port, possibly on VLAN `x`.
  * Strip VLAN, Geneve encapsulate packet using fields in `group_entry.nat_target`.
  * Adjust TTL.
  * Replicate Geneve encapsulated frames out on all rear ports on the mcast group.
* Encapsulated multicast packet arrives on a rear panel port.
  * Adjust TTL.
  * Replicate frame, including Geneve, onto all _other_ rear panel ports in the group.
  * Replicate frame, excluding Geneve, onto all front panel ports in the group. Insert VLAN if required by link.

Am I coming at this the wrong way?

FelixMcFelix (commented):
I think I've had a look over all the changes now, thanks for acting on the last round of feedback.

dpd-client/tests/integration_tests/mcast.rs
zeeshanlakhani: Yeah, I get your question (and we're chatting tmrw); however, I'm trying to map/orient how we should do this (vs unicast). Today, `nat_egress` is triggered when we match a switch_address (a v4/v6 address assigned to ports on the switch when configuring port settings on switch initialization). I definitely don't know enough (classic!), but for multicast group creation, there's nothing specifically tying multicast groups to ingress ports (so that we'd know if it's a front vs rear panel port or not), and it will dynamically change. 

So, we'd need to explore an alternative approach to what we did for unicast before, though we could reuse the switch tables, possibly?

dpd/src/table/route_ipv4.rs
```diff
@@ -30,7 +30,7 @@ struct IndexKey {
 
 // Route entries stored in the index->route_data table
 #[derive(ActionParse, Debug)]
-enum RouteAction {
+pub(crate) enum RouteAction {
```
Nieuwejaar, line 33: I know it leads to a little code duplication, but I'd like to keep the table keys private to their modules.

dpd/src/types.rs
```diff
@@ -85,6 +85,10 @@ pub enum DpdError {
     Oximeter(String),
     #[error("No switch identifiers available")]
     NoSwitchIdentifiers,
+    #[error("Multicast group error: {}", .0)]
+    McastGroup(String),
+    #[error("Resource exhausted: {}", .0)]
```
Nieuwejaar, line 90: The only place this is used is when we're out of group IDs.  You could just make that a specific error, rather than the vague "resource exhausted".

packet/src/ipv4.rs
```diff
@@ -104,6 +104,19 @@ impl Ipv4Hdr {
         hdr.ipv4_ttl = ttl as u8;
         Ipv4Hdr::update_checksum(pkt);
     }
+
+    pub fn set_identification(pkt: &mut Packet, id: u16) {
```
Nieuwejaar, line 108: Neither of these seem to be used anywhere.

packet/src/ipv6.rs
```diff
@@ -62,9 +62,9 @@ impl Ipv6Hdr {
         v
     }
 
-    pub fn adjust_hlim(pkt: &mut Packet, delta: i8) {
-        let hdr = pkt.hdrs.ipv6_hdr.as_mut().unwrap();
-        hdr.ipv6_hop_lim = (hdr.ipv6_hop_lim as i8 + delta) as u8;
+    pub fn adjust_hlim(pkt: &mut Packet, delta: i16) {
```
Nieuwejaar, line 65: This is an 8 bit field - why allow a 16 bit delta?

dpd/p4/constants.p4
```diff
@@ -1,36 +1,69 @@
...
-const int SWITCH_IPV4_ADDRS_SIZE	= 512;  // ipv4 addrs assigned to our ports
-const int SWITCH_IPV6_ADDRS_SIZE	= 512;  // ipv6 addrs assigned to our ports
-
-const bit<8> SC_FWD_FROM_USERSPACE	= 0x00;
-const bit<8> SC_FWD_TO_USERSPACE	= 0x01;
-const bit<8> SC_ICMP_NEEDED		= 0x02;
-const bit<8> SC_ARP_NEEDED		= 0x03;
-const bit<8> SC_NEIGHBOR_NEEDED		= 0x04;
-const bit<8> SC_INVALID			= 0xff;
+// Includes the checksum for the original data, the geneve header, the
```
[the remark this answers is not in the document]
zeeshanlakhani: Ok, yeah, I'll move defines to the top of the sidecar.

dpd/src/api_server.rs
```diff
@@ -3061,6 +3066,260 @@ async fn counter_get(
...
+    query_params: Query<
+        PaginationParams<EmptyScanParams, MulticastGroupIpParam>,
+    >,
+) -> Result<HttpResponseOk<ResultsPage<mcast::MulticastGroupResponse>>, HttpError>
+{
+    let switch: &Switch = rqctx.context();
+    let group_id = query_id.into_inner().group_id;
+
+    // If a group ID is provided, get the group by ID
+    if let Some(group_id) = group_id {
```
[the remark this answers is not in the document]
zeeshanlakhani: yeah, that's why I made this a query param if you want to use the ID (vs the {ip} variant). But, maybe I should just remove the get by group_id entirely to avoid confusion.

dpd/src/route.rs
```diff
@@ -153,7 +154,18 @@ impl<T: Clone> RouteEntry<T> {
...
-#[derive(Debug, Eq, PartialEq, Ord, PartialOrd, Clone)]
+/// A Vlan identifier, made up of a port, link, and vlan tag.
+#[derive(
+    Debug,
+    Eq,
+    PartialEq,
+    Ord,
+    PartialOrd,
+    Clone,
+    Deserialize,
```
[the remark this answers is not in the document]
zeeshanlakhani: I waas using it previously, but dropped it. Good catch, will remove.

dpd/src/table/mcast.rs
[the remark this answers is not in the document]
zeeshanlakhani: Yeah, I'll move to the "older style".

dpd/p4/constants.p4
```diff
@@ -1,36 +1,69 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
```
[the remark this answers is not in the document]
zeeshanlakhani: I'll double-check this.

packet/src/ipv4.rs
zeeshanlakhani: Yep. I was initially setting these into the packet to copy over some mcast info. But, I'll remove this.

packet/src/ipv6.rs
zeeshanlakhani: It was so I could update it to set it to 0 for a test:

> ipv6::Ipv6Hdr::adjust_hlim(&mut to_send, -255);

dpd/p4/constants.p4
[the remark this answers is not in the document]
zeeshanlakhani: I did make changes post @FelixMcFelix and @rcgoodfellow's review, so maybe that's why?

dpd/src/api_server.rs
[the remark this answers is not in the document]
Nieuwejaar: I think I'm saying the opposite of that.  I'm asking you to add a second entry point rather than use a query param, so this is consistent with the other APIs.  If you don't think there is a use case for getting a single group, then you can remove the query param and not add a second endpoint.

dpd/src/types.rs
zeeshanlakhani: ResourceExhaused is for the next line, but I'll rename this to something more apt.

dpd-client/tests/integration_tests/mcast.rs
FelixMcFelix: >     * Encapsulated multicast packet arrives on a rear panel port.
>       
>       * Adjust TTL.
>       * Replicate frame, including Geneve, onto all _other_ rear panel ports in the group.
>       * Replicate frame, excluding Geneve, onto all front panel ports in the group. Insert VLAN `x` if required by link.

We talked over this one a bit, there are some thorny parts around source address NAT'ing in this pathway. One way we were looking at this was to make use of both multicast group IDs available to the ASIC. I.e., we'd replicate one set of packets to _rear ports in the group_, and another to _front ports_ (stripping encap and pushing VLAN). But then we've either applied NAT on the source addr of rack-local packets, or left the private source address in place for rack egress.

I *think* our disposition is that OPTE needs to (when required) produce two copies of each northbound packet itself:
* One uses the guest's VNI using whatever multicast v6 address we bind to the group. NAT is not applied, and this packet is sent almost verbatim to the underlay half of the replication group.
* One targets the switch address (or a dedicated m'cast switch address) as with V2B mappings today. This packet has NAT applied in OPTE, and is decapped and replicated to all rack-external recipients.

There might be other parts to NAT that we're missing here.

FelixMcFelix: > I think our disposition is that OPTE needs to (when required) produce two copies of each northbound packet itself:

Following up on this, we don't need to generate two multicast variants of a given packet sent by a guest. From discussion yesterday, the plan is that all multicast packets sent by guests will use an external IP address for their source even when delivered internal to the VPC. Thus, no strange head-end replication.

dpd/p4/sidecar.p4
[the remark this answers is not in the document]
zeeshanlakhani: New version changes all of this.

zeeshanlakhani:
> Got this back down to 14 stages.

With new additions, we are at 18 (comments inside the sidecar).

dpd-client/tests/integration_tests/mcast.rs
```diff
@@ -1205,1177 +1276,2557 @@ async fn test_ipv6_multicast_basic_replication() -> TestResult {
...
+    let expected_pkts = vec![
+        TestPacket {
+            packet: Arc::new(to_recv1),
+            port: egress1,
+        },
+        TestPacket {
+            packet: Arc::new(to_recv2),
+            port: egress2,
+        },
+    ];
```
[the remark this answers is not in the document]
zeeshanlakhani: Yeah, let's chat in the morning (for me haha). It's more than this is possible depending on the traffic coming into the switch. We could not allow it?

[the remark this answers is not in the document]
zeeshanlakhani: Ok, post-discussion, going to make updates to where we don't allow external to external replication (i.e. translation).

FelixMcFelix (commented):
Thanks Zeeshan, I've had a look mainly focussing on the new changes and `dpd/drc/mcast/*.rs`. I do like the removal of v4 multicast from the routing table, and then relying on the NAT target to make use of it from external ports.

dpd/src/mcast/mod.rs
```diff
@@ -203,13 +227,13 @@ impl MulticastGroupData {
...
             groups: BTreeMap::new(),
-            // Start at a threshold to avoid early allocations
             id_generator: AtomicU16::new(Self::GENERATOR_START),
             used_group_ids: HashSet::new(),
+            nat_target_refs: BTreeMap::new(),
         }
     }
 
-    /// Generate a unique multicast group ID.
+    /// Generates a unique multicast group ID by finding the next available ID.
```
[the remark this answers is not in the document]
zeeshanlakhani: I can make this better.

dpd/src/mcast/mod.rs
```diff
@@ -99,40 +104,59 @@ pub(crate) struct MulticastGroup {
...
     pub(crate) sources: Option<Vec<IpSrc>>,
-    pub(crate) replication_info: MulticastReplicationInfo,
+    pub(crate) replication_info: Option<MulticastReplicationInfo>,
     pub(crate) members: Vec<MulticastGroupMember>,
 }
 
-/// A multicast group entry for POST requests.
-#[derive(Debug, Deserialize, Serialize, JsonSchema)]
+/// Represents a multicast replication entry for POST requests.
+#[derive(Debug, Default, Deserialize, Serialize, JsonSchema)]
```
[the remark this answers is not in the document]
zeeshanlakhani: Yeah. I see what you're saying. I'll remove that.

FelixMcFelix (commented):
Thanks for being patient. I think I'm happy with the API shape, just some points on cleanup and lock hygiene.

zeeshanlakhani:
@FelixMcFelix should now be in a good spot API-wise.

dpd/src/mcast/mod.rs
```diff
@@ -1333,15 +1407,19 @@ fn ensure_external_group_exists(
...
 ) -> DpdResult<()> {
     if group_entry.external_group_id.is_none()
         && members.iter().any(|m| m.direction == Direction::External)
     {
-        let group_id = mcast.generate_group_id()?;
-        create_asic_group(s, group_id, group_ip)?;
+        let scoped_group_id = pre_allocated_id.ok_or_else(|| {
+            DpdError::Invalid(
+                "external group ID should have been pre-allocated".to_string(),
+            )
```
FelixMcFelix, line 1418: This feels like an invariant maintained server-side, rather than something derived from poor user input. `DpdError::Invalid` will convert into an HTTP 400, whereas this seems like a 500 on that note. `DpdError::Other` might fill that niche?

zeeshanlakhani:
Pinging @Nieuwejaar and @rcgoodfellow for any other looks here beyond @FelixMcFelix's extensive review!

rcgoodfellow:
Unless @Nieuwejaar has any more feedback I'd say let's get this landed.

.gitignore
```diff
@@ -17,6 +17,7 @@ p4_artifacts*
 # Editor config
 .vscode
 .dir-locals.el
+bacon.toml
```
Nieuwejaar, line 20: Should this be removed?

dpd/p4/constants.p4
```diff
@@ -1,36 +1,43 @@
...
+// is going to need to come either through ATCAM/ALPM or code restructuring.
+const int IPV4_NAT_TABLE_SIZE       = 1024; // nat routing table
+const int IPV6_NAT_TABLE_SIZE       = 1024; // nat routing table
+const int IPV4_LPM_SIZE             = 8192; // ipv4 forwarding table
+const int IPV6_LPM_SIZE             = 1024; // ipv6 forwarding table
+const int IPV4_ARP_SIZE             = 512;  // arp cache
+const int IPV6_NEIGHBOR_SIZE        = 512;  // ipv6 neighbor cache
+const int SWITCH_IPV4_ADDRS_SIZE    = 512;  // ipv4 addrs assigned to our ports
+const int SWITCH_IPV6_ADDRS_SIZE    = 512;  // ipv6 addrs assigned to our ports
+const int IPV4_MULTICAST_TABLE_SIZE = 1024; // multicast routing table(s) for IPv4
```
Nieuwejaar, line 19: How did you arrive at a size of 1024 for all of the multicast tables?  I suspect that when we add customer support for IPv6, we're going to have to bump up the existing IPv6 tables, and we don't have a lot of headroom.  Unless we know that we need the multicast tables to be this big, I would be inclined to start with something much smaller and grow them if the customer hits the limit.  That will be much lower risk than trying to shrink them after this is in use at a customer site.

.gitignore
zeeshanlakhani: I had been using it (and it is a tool, but find to remove it).

dpd/src/api_server.rs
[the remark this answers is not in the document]
zeeshanlakhani: > Is there a reason we can't have /multicast/groups/{group_id}? We already have a MulticastGroupIdParam::to_identifier() operation that could convert it from either format.

I had something like this in a previous iteration, but dropshot wouldn't allow it: https://github.com/oxidecomputer/dendrite/pull/14#discussion_r2051970521.

dpd/p4/constants.p4
zeeshanlakhani: @Nieuwejaar I thought 1024 was starting small, based on what mcast-interested customers had been bringing up (which was larger). I could break up the V6 constants as they're used for different table components, but maybe @rcgoodfellow has thoughts here.

.gitignore
Nieuwejaar: It's harmless.  I just wasn't sure if you meant to leave it in or not.  If you want to keep it, feel free.

PR #20, opened 2025-04-08
Report bit-error rates from links with RS FEC enabled

bnaecker:
- Add ASIC-layer methods for reporting the BER from an ASIC port
- Export Tofino2 functions and types for reading BER out of the PHY
- Add `dpd` and `swadm` interfaces for extracting the BER and printing it

### DIFF
--- a/asic/src/tofino_asic/ports.rs
+++ b/asic/src/tofino_asic/ports.rs
@@ -41,6 +41,7 @@ use std::fmt;
 use slog::debug;
 
 use aal::AsicId;
+use aal::Ber;
 use aal::Connector;
 use aal::PortHdl;
 
@@ -189,7 +190,7 @@ impl fmt::Debug for FrontPortHandle {
     }
 }
 
-fn speed_to_bf(speed: PortSpeed) -> bf_port_speed_t {
+fn port_speed_to_bf_speed(speed: PortSpeed) -> bf_port_speed_t {
     match speed {
         PortSpeed::Speed0G => bf_port_speed_e_BF_SPEED_NONE,
         PortSpeed::Speed1G => bf_port_speed_e_BF_SPEED_1G,
@@ -200,6 +201,21 @@ fn speed_to_bf(speed: PortSpeed) -> bf_port_speed_t {
         PortSpeed::Speed100G => bf_port_speed_e_BF_SPEED_100G,
         PortSpeed::Speed200G => bf_port_speed_e_BF_SPEED_200G,
         PortSpeed::Speed400G => bf_port_speed_e_BF_SPEED_400G,
+    }
+}
+
+fn bf_speed_to_port_speed(speed: bf_port_speed_t) -> PortSpeed {
+    match speed {
+        bf_port_speed_e_BF_SPEED_NONE => PortSpeed::Speed0G,
+        bf_port_speed_e_BF_SPEED_1G => PortSpeed::Speed1G,
+        bf_port_speed_e_BF_SPEED_10G => PortSpeed::Speed10G,
+        bf_port_speed_e_BF_SPEED_25G => PortSpeed::Speed25G,
+        bf_port_speed_e_BF_SPEED_40G => PortSpeed::Speed40G,
+        bf_port_speed_e_BF_SPEED_50G => PortSpeed::Speed50G,
+        bf_port_speed_e_BF_SPEED_100G => PortSpeed::Speed100G,
+        bf_port_speed_e_BF_SPEED_200G => PortSpeed::Speed200G,
+        bf_port_speed_e_BF_SPEED_400G => PortSpeed::Speed400G,
+        _ => unreachable!("Impossible bf_port_speed_t"),
     }
 }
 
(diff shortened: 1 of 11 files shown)


asic/src/tofino_asic/ports.rs
```diff
@@ -412,6 +413,45 @@ pub fn set_prbs(
...
+        }
+        e => {
+            slog::error!(hdl.log, "failed to get port BER: {:?}", e);
+            return Err(sde_error("getting port BER", e));
+        }
+    };
+    Ok(Ber {
+        symbol_errors: ber.sym_err_ctrs.to_vec(),
+        ber: ber.ber_per_lane.to_vec(),
+        total_ber: ber.ber_aggr,
```
bnaecker, line 451: I'd prefer to extract only the counters that are specific to the lanes this link uses, but the SDE does not make it easy to understand which those are. It relies on some bit-twiddling that's done [here](https://github.com/oxidecomputer/tofino-sde/blob/0e4cfe10c92db223a953923d77f09a299b82157b/pkgsrc/bf-drivers/src/bf_pm/pm.c#L2120-L2140) -- it's not super complicated, but it also feels a bit fraught to reimplement this in the ASIC layer. I can do that if we think it's valuable, though.

bnaecker:
I tested this on **london**, by creating two links on `qsfp3` and `qsfp4`. The former has 100G FR1 transceivers, which internally implement FEC and so don't support the BER API that the SDE provides. The latter does support it, though the counters are all zero right now, even with a PRBS pattern.

```
root@oxz_switch0:~# swadm link counters ber qsfp4/0
Lane   BER      Symbol errors
0      0.000e0  0
1      0.000e0  0
2      0.000e0  0
3      0.000e0  0
4      0.000e0  0
5      0.000e0  0
6      0.000e0  0
7      0.000e0  0
8      0.000e0  0
9      0.000e0  0
10     0.000e0  0
11     0.000e0  0
12     0.000e0  0
13     0.000e0  0
14     0.000e0  0
15     0.000e0  0
Total  0.000e0  0
root@oxz_switch0:~# swadm link counters ber qsfp3/0
Error: failed to fetch link BER

Caused by:
    0: failed to get link BER
    1: Error Response: status: 400 Bad Request; headers: {"content-type": "application/json", "x-request-id": "b2c74897-cfe2-4b87-aecc-0dc12589fba2", "content-length": "129", "date": "Tue, 08 Apr 2025 21:39:54 GMT"}; value: Error { error_code: None, message: "BER can only be computed for links with RS FEC enabled", request_id: "b2c74897-cfe2-4b87-aecc-0dc12589fba2" }
```

bnaecker:
This is intended to resolve https://github.com/oxidecomputer/dendrite-archived/issues/1105

asic/src/tofino_asic/ports.rs
bnaecker: I added this in https://github.com/oxidecomputer/dendrite/pull/20/commits/f170ce7fdcfb1aaed02ac8d6394c4672e331bdfb. I'm ambivalent about it, since we need to make additional calls into the SDE to get the speed of the ASIC port, and then duplicate some opaque logic that's in the SDE to figure out the right counters. But it's much more helpful to report only the counters that _should_ matter for a link, so I've included it for now.

asic/src/tofino_asic/ports.rs
```diff
@@ -412,6 +428,118 @@ pub fn set_prbs(
...
+            return Err(sde_error("getting port BER", e));
+        }
+    };
+
+    // Fetch the port speed, which is needed to determine which counters from
+    // the above call to `bf_pm_port_ber_get` are valid for this port.
+    let bf_speed = bf_port_speed_e_BF_SPEED_NONE;
+    let dev_port = fp.get_dev_port()?.into();
+    let res =
+        unsafe { bf_port_speed_get(dev, dev_port, &bf_speed as *const _ as _) };
```
Nieuwejaar, line 525: Don't we already know the port speed?

asic/src/tofino_asic/ports.rs
```diff
@@ -412,6 +428,118 @@ pub fn set_prbs(
...
+        // This magic logic appears in `pm_port_tof2_ber_get()`, where it
+        // appears to be used to determine the right counters to read from the
+        // serdes firmware.
+        const MASK: u64 = 0x07;
+        const FACTOR: u64 = 2;
+        let Ok(start) = usize::try_from((u64::from(hdl) & MASK) * FACTOR)
+        else {
+            return Self::all();
+        };
+        match speed {
```
Nieuwejaar, line 476: This calculation doesn't seem complete.  Can't we build a 100g port from 10 10g links?  40g from 4 10g links?  There is a `serdes::lane_count()` function that will return the actual number of lanes configured.

asic/src/tofino_asic/ports.rs
bnaecker: Not really, at this layer of the code anyway. We could pass it down in the interface to the new BER trait method, it just seemed less helpful of an interface. I don't have a strong preference, we can do it that way too.

asic/src/tofino_asic/ports.rs
bnaecker: I agree that you _could_, but I'm not sure we support that today. Don't we or the SDE automatically figure out the number of lanes? That's what `PhysPort::allocate_channels_for()` is doing, anyway.

I'm also trying to replicate exactly what the SDE does internally, since I have very little insight into what the various counters actually mean. The other option is to ignore all this, and return 16 of them all the time. I dislike that a bit because then it's on the operator to figure out if zero means "this is fine" or "this wasn't used".

bnaecker: I think I see a better way to do this, in any case, and in a way that avoids needing the speed at all. We know the number of _lanes_, as you point out, and we can replicate the logic in the SDE with just ASIC port identifier. I.e., we don't need the speed at all. I'll rework all this.

asic/src/tofino_asic/ports.rs
bnaecker: Obviated by [5ce22d7](https://github.com/oxidecomputer/dendrite/pull/20/commits/5ce22d7c672cf3fcffce5bda96cb4f2871d766f6)

asic/src/tofino_asic/ports.rs
bnaecker: Done in [5ce22d7](https://github.com/oxidecomputer/dendrite/pull/20/commits/5ce22d7c672cf3fcffce5bda96cb4f2871d766f6)

Nieuwejaar (approved):

PR #22, opened 2025-04-14
Notify Nexus about the switch Dendrite manages on startup

bnaecker:


bnaecker:
Talked offline with @davepacheco, I'm going to close this and pursue fetching this information from Nexus instead.

PR #48, opened 2025-06-11
hardware-cosmo#601 swadm compliance subcommand

taspelund:


swadm/src/compliance.rs
```diff
@@ -0,0 +1,275 @@
...
+    let mut created_count = 0;
+    let mut error_count = 0;
+
+    for port_id in switch_ports {
+        let params = types::LinkCreate {
+            lane: None, // Use default first lane
+            speed: speed.into(),
+            fec: Some(fec.into()),
+            autoneg,
+            kr: false, // Default to false for compliance
```
rcgoodfellow, line 249: If `--all` is specified, this will be incorrect for backplane links.

swadm/src/compliance.rs
```diff
@@ -0,0 +1,275 @@
...
+        switch_ports.len(), port_type, speed, fec, autoneg);
+
+    let mut created_count = 0;
+    let mut error_count = 0;
+
+    for port_id in switch_ports {
+        let params = types::LinkCreate {
+            lane: None, // Use default first lane
+            speed: speed.into(),
+            fec: Some(fec.into()),
```
rcgoodfellow, line 247: Per earlier comment I think we should just specify `None` here.

swadm/src/compliance.rs
```diff
@@ -0,0 +1,275 @@
...
+) -> anyhow::Result<()> {
+    // Get all switch ports
+    let all_ports = client
+        .port_list()
+        .await
+        .context("failed to list switch ports")?
+        .into_inner();
+
+    // Filter to qsfp ports only unless --all is specified
+    let switch_ports: Vec<types::PortId> = if all {
```
rcgoodfellow, line 227: I believe dendrite automatically creates backplane links. I can see this being useful if we need to delete and then recreate backplane links for compliance, but I don't see the corresponding link deletion code?

taspelund: Dendrite does automatically create backplane links.  I didn't think we needed/wanted to be able to delete links, just disable them.  Happy to add a corresponding teardown command though with the same filtering semantics

taspelund: I'd like to keep the `--all` flag since I want rfk to be able to manage the backplane links just in case the need arises.  I don't anticipate he'll need to, but the tool should allow for it IMO

swadm/src/compliance.rs
```diff
@@ -0,0 +1,613 @@
...
+        qsfp_ports
+            .into_iter()
+            .filter(|port| port.to_string().contains(user_pattern))
+            .collect()
+    } else {
+        qsfp_ports
+    };
+
+    if target_ports.is_empty() {
+        println!("No qsfp ports found matching pattern");
```
Nieuwejaar, line 495: `eprintln`?

swadm/src/compliance.rs
```diff
@@ -0,0 +1,613 @@
...
+
+async fn compliance_ports_enable(
+    client: &Client,
+    pattern: Option<&str>,
+    enabled: bool,
+    all: bool,
+) -> anyhow::Result<()> {
+    let links = get_matching_links(client, pattern, all).await?;
+
+    for link in links {
```
Nieuwejaar, line 282: Not a big deal, but I think this could be tightened up a bit:
```
    let op = if enabled { "enable" } else { "disable" };
    for link in links {
        client
            .link_enabled_set(&link.port_id, &link.link_id, enabled)
            .await
            .context(format!("failed to {op} link {link}",))?;
        println!("{op}d link {link}");
    }

Nieuwejaar (approved):

PR #82, opened 2025-07-30
Add MSS Geneve option, support several Geneve options

FelixMcFelix:
This PR retools our parsing of Geneve options to rely upon a `ParserCounter`. This allows us to keep track of how many 4B chunks within the header are remaining after successfully parsing each option, and to reject packets with overly long option lists.

Limitations:
* Output packets will have all their Geneve options reordered, since we do not make use of a header stack.
* We support at most one of each Geneve option, which is probably reasonable enough.
* We don't have a mechanism for transit of *unknown* Geneve options, so no forward-compatibility yet.

See also https://github.com/oxidecomputer/sidecar-lite/pull/79, https://github.com/oxidecomputer/opte/pull/805.

### DIFF
--- a/.github/buildomat/jobs/image.sh
+++ b/.github/buildomat/jobs/image.sh
@@ -101,7 +101,7 @@ pfexec chown "$UID" /out
 
 banner "P4 Codegen"
 # Add gcc-12 so the p4 compiler can find cpp
-# The tofino2 has 20 stages, but the current sidecar.p4 will fit into 18.  We
+# The tofino2 has 20 stages, but the current sidecar.p4 will fit into 19.  We
 # add the "--stages 19" here to detect if/when the program grows beyond that
 # limit.  It's not necessarily a problem if we grow, but given the limited space
 # on the ASIC, we want to grow deliberatately and thoughtfully.
--- a/.github/buildomat/jobs/image.sh
+++ b/.github/buildomat/jobs/image.sh
@@ -102,10 +102,10 @@ pfexec chown "$UID" /out
 banner "P4 Codegen"
 # Add gcc-12 so the p4 compiler can find cpp
 # The tofino2 has 20 stages, but the current sidecar.p4 will fit into 18.  We
-# add the "--stages 18" here to detect if/when the program grows beyond that
+# add the "--stages 19" here to detect if/when the program grows beyond that
 # limit.  It's not necessarily a problem if we grow, but given the limited space
 # on the ASIC, we want to grow deliberatately and thoughtfully.
-PATH=/opt/gcc-12/bin:$PATH cargo xtask codegen --stages 18
+PATH=/opt/gcc-12/bin:$PATH cargo xtask codegen --stages 19
 
 # Preserve all the diagnostics spit out by the compiler
 mkdir -p /out/p4c-diags
--- a/dpd-client/tests/integration_tests/common.rs
+++ b/dpd-client/tests/integration_tests/common.rs
@@ -929,6 +929,153 @@ pub fn gen_udp_routed_pair(
     gen_udp_routed_pair_loaded(switch, phys_port, router, src, dst, &Vec::new())
 }
 
+#[derive(Copy, Clone, Debug, Hash, Eq, PartialEq, Ord, PartialOrd)]
+pub enum OxideGeneveOption {
+    External,
+    Multicast(u8),
+    Mss(u32),
+}
+
+/// Build a Geneve packet with a possible multicast tag.
+pub fn gen_geneve_packet_with_mcast_tag(
+    src: Endpoint,
+    dst: Endpoint,
+    inner_type: u16,
+    vni: u32,
+    mcast_tag: u8,
+    payload: &[u8],
+) -> Packet {
+    let opt = OxideGeneveOption::Multicast(mcast_tag);
+    gen_geneve_packet(
+        src,
+        dst,
+        inner_type,
+        vni,
+        core::slice::from_ref(&opt),
+        payload,
+    )
+}
+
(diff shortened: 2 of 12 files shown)


.github/buildomat/jobs/image.sh
```diff
@@ -101,11 +101,11 @@ pfexec chown "$UID" /out
 
 banner "P4 Codegen"
 # Add gcc-12 so the p4 compiler can find cpp
-# The tofino2 has 20 stages, but the current sidecar.p4 will fit into 18.  We
-# add the "--stages 18" here to detect if/when the program grows beyond that
+# The tofino2 has 20 stages, but the current sidecar.p4 will fit into 19.  We
```
Nieuwejaar, line 104: yikes.

dpd-client/tests/integration_tests/common.rs
```diff
@@ -929,6 +929,153 @@ pub fn gen_udp_routed_pair(
...
+                panic!("illegal specification for option: {opt:?}")
+            }
+        }
+    }
+
+    let extra_bytes = geneve.options.len() as u16;
+    assert_eq!(
+        extra_bytes % 4,
+        0,
+        "Geneve options must be pushed in 4B blocks, found {extra_bytes}",
```
Nieuwejaar, line 1058: Huh.  I didn't know `assert` let you provide an explanation.  That's very handy.

dpd/p4/parser.p4
```diff
@@ -364,54 +367,93 @@ parser IngressParser(
...
 		}
 	}
 
 	state parse_geneve_ox_opt {
-		transition select(hdr.geneve_opts.ox_external_tag.type) {
-			GENEVE_OPT_OXIDE_EXTERNAL: geneve_parsed;
+		bit<24> curr_opt = pkt.lookahead<bit<24>>();
+		geneve_chunks.decrement(1);
+
+		transition select(curr_opt[6:0]) {
```
Nieuwejaar, line 390: It seems like if we have two headers with the same option, the second header will overwrite the first. Is that right?  That seems reasonable to me, unless the standard/RFC says that it should be an error.

dpd/p4/parser.p4
```diff
@@ -364,54 +367,93 @@ parser IngressParser(
...
-		//    + This allows varbits outside of header stacks.
-		//    + No need for Tofino externs.
-		//    - 4B overhead/pkt iff. other options.
-		// 2) Use a ParserCounter.
-		//    + Better validation/rejection of bad opt_lens.
-		//    + No per-packet overhead.
-		//    - ICRP forums suggest higher parse cost?
-		//    - Probably a lot of ugly states/branching on opt_len
-		//      to get a const value for counter decrement.
+		geneve_chunks.set(hdr.geneve.opt_len);
```
Nieuwejaar, line 370: Is there a limit to how many times we can iterate on this?  I'm not sure why they would require us to use the counter if there wasn't.  This really just curiosity on my part - not a request to change anything.

dpd/p4/sidecar.p4
```diff
@@ -1951,6 +1958,31 @@ control Ingress(
...
+			//   parse (nat_ingress).
+			//
+			// - Why are we storing this in metadata?
+			//   > error: Non-zero constant entry in checksum
+			//     calculation not implemented yet: 16w0x129
+			//   Fairly straightforward, and I imagine this could
+			//   be the easier one to fix in tofino-p4c. This
+			//   `todo!()` covers various other tricks, including
+			//   wrapping the const in a struct/header.
+			meta.nat_ingress_csum = 16w0x0129;
```
Nieuwejaar, line 1984: It's not obvious to me why this is always 0x129.  I believe you that it is, but I could use an explanation here.

Nieuwejaar (commented):
I have  few questions, but overall this is a really clever bit of work - especially in the p4 code.

dpd/p4/parser.p4
FelixMcFelix: As written, that's how this will work. I don't see any case yet where OPTE or Dendrite would push multiple copies of a single option unless we needed to shift more than 124B of payload for the same option type, but you never know.

The [RFC](https://datatracker.ietf.org/doc/html/rfc8926) is more stringent about what transit should offer when a node interprets Geneve and inner headers:
* §3.5.1 -- 'contents of the options and their ordering MUST NOT be modified by transit devices'.
* §3.5.1 -- 'transit devices MUST NOT drop packets as a result of encountering unknown options, including those with the 'C' bit set'.
* §4.5.1 -- 'a control plane may restrict the number of option TLVs as well as the order and size of the TLVs between tunnel endpoints'.

So we're failing on the two 'MUST NOT' points, but we weren't succeeding at them before. I guess this is technically a restriction on option ordering and size/count.

To provide these guarantees, we'd likely need to make use of header stacks (which we have been [encouraged not to use](https://github.com/oxidecomputer/sidecar-lite/blob/3a5ae33098b92ade26131b40939e27578a0ec5e4/p4/headers.p4#L144-L146)). I think that puts us in a slightly strange spot, since we seemingly can't use `header union`s and we can't extract/push when we have a `varbit` in a header stack. So we might need to represent that as a stack of opaque 4-byte blocks, including the option headers. We could probably assert that any 'switch-relevant' options (like multicast info) come first, so we can make use of them. We could try and prototype something here (and I'd like to properly support this).

Nieuwejaar: This handling is all very specific to our use case, so I guess I wouldn't get too worked up about sticking with the RFC.  Even in a multirack scenario, where our traffic could be transiting a third party network, how we handle our own geneve options in our switch code shouldn't be an issue.

Nieuwejaar (approved):

PR #88, opened 2025-08-07
dendrite#87 all tables should have counters added

Nieuwejaar:
Adds table counters to the 2 mac rewrite tables.
Adds new counters to the Service control, so each path is counted - including "take no action".
Adds a counter for unicast packets in the Egress pipeline
dpd was returning the `pipe.Egress.mcast_ctr` counter for all multicast counter requests from `swadm`.  Added new `CounterId` variants to catch them all.

### DIFF
--- a/dpd/p4/sidecar.p4
+++ b/dpd/p4/sidecar.p4
@@ -292,7 +292,6 @@ control Services(
 	// It's not strictly necessary, but we also bump up the TTL to make sure
 	// the packet makes it all the way back to the sender.
 	action ping4_reply() {
-		service_ctr.count(SVC_COUNTER_V4_PING_REPLY);
 		hdr.ethernet.dst_mac = meta.orig_src_mac;
 
 		hdr.ipv4.src_addr = hdr.ipv4.dst_addr;
@@ -305,10 +304,10 @@ control Services(
 		ig_tm_md.ucast_egress_port = ig_intr_md.ingress_port;
 
 		meta.service_routed = true;
+		service_ctr.count(SVC_COUNTER_V4_PING_REPLY);
 	}
 
 	action ping6_reply() {
-		service_ctr.count(SVC_COUNTER_V6_PING_REPLY);
 		hdr.ethernet.dst_mac = meta.orig_src_mac;
 
 		bit<128> orig_src = hdr.ipv6.src_addr;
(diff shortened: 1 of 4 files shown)


dpd/p4/sidecar.p4
```diff
@@ -2130,7 +2147,7 @@ control Egress(
 		} else if (is_mcast == true) {
 			mcast_ctr.count(eg_intr_md.egress_port);
 
-			if (is_ipv6_mcast) {
+			if (!is_ipv6_mcast) {
```
Nieuwejaar, line 2150: At line 2122 we have:
```
is_ipv6_mcast = (ipv6_prefix != 16w0xff02);
```
`ff02` is for link-local multicast, so `is_ipv6_mcast` means any non-linklocal multicast, right?  So we want to bump the link-local counter for `!is_ipv6_mcast`?  Or am I getting turned around by all the `not`s?

bnaecker: The logic or possibly the variable naming seems kind of confused here, before this change. `0xff` is the multi-cast prefix, and `0x02` means "link-local scope". So `is_ipv6_mcast` is true for anything that is not _exactly_ a link-local multicast address. It is true for multicast addresses of other scopes, or for addresses that aren't multicast at all like link-local addresses or globally-unique addresses.

What are we trying to catch here? _Only_ link-local multicast addresses? If that's true, then I'd probably invert the prefix check on L2122 and rename the variable to `is_ipv6_ll_mcast` or something like that.

zeeshanlakhani: @Nieuwejaar yeah. My bad. I think this got lost in the review changes. This should be renamed to `is_link_local_mcastv6` and be   

```
is_link_local_ipv6_mcast = (ipv6_prefix == 16w0xff02);
```

`egress_rid` will capture replicated mcast packets. So, I think it's renamed for the counter, and we don't need the `!`.

zeeshanlakhani: > The logic or possibly the variable naming seems kind of confused here, before this change. `0xff` is the multi-cast prefix, and `0x02` means "link-local scope". So `is_ipv6_mcast` is true for anything that is not _exactly_ a link-local multicast address. It is true for multicast addresses of other scopes, or for addresses that aren't multicast at all like link-local addresses or globally-unique addresses.
> 
> What are we trying to catch here? _Only_ link-local multicast addresses? If that's true, then I'd probably invert the prefix check on L2122 and rename the variable to `is_ipv6_ll_mcast` or something like that.

Yeah. What I said in my comment (before I expanded yours @bnaecker) haha.

Nieuwejaar: I made the suggested change to `is_link_local_ipv6_mcast`

zeeshanlakhani (approved):

PR #117, opened 2025-09-04
remove listen_addr requirement for rpw startup

internet-diglett:


rcgoodfellow:
Initially I was a bit concerned about how this would interact with early networking. As this will result in the RPW starting up sooner. But it looks like we assign a non-localhost address in `boot_switch_zone` in sled agent, so we're already getting past this point by the time early networking config is being set up by sled agent. Is that your read @internet-diglett? I feel like we should take a racklette run with an omicron branch that pulls this commit in.

rcgoodfellow (approved):

PR #129, opened 2025-09-14
softnpu: fix router v6 keyset data

rcgoodfellow:


asic/src/softnpu/table.rs
```diff
@@ -564,6 +564,11 @@ fn keyset_data(match_data: Vec<MatchEntryField>, table: &str) -> Vec<u8> {
                         serialize_value_type(&x, &mut data);
                         keyset_data.extend_from_slice(&data[..2]);
                     }
+                    ROUTER_V6_RT => {
```
Nieuwejaar, line 567: Huh.  How did I get this up on a4x2 if ipv6 routing doesn't work?

rcgoodfellow: This is not the cause of the issue in oxidecomputer/maghemite#545. I can reproduce the issue with this patch.

Issue #104, opened 2025-08-27
`port_settings_apply` does not apply tx_eq settings


rcgoodfellow:
In early networking before nexus starts, we depend on sled agent to configure the switch for external networking. Sled agent uses `port_settings_apply`. That function does not apply tx equlization settings, either from the bootstore config, or attempt to use the default value for a given transceiver. This results in sadness for transceivers that require something different than the SDE defaults.

We have several different code paths that update link settings. There is `port_settings_apply`, the reconciler/workflow server that runs in dpd and the API functions that `swadm` uses. We should try to consolidate these.

rcgoodfellow:
It looks like `port_settings_apply` now emits events that cause parts of the reconciler to run. Which should include setting tx equalization settings. So these code paths are not as disparate as I thought.

rcgoodfellow:
Fixed in 
- #137

PR #137, opened 2025-09-24
do not reconcile link settings if transceiver is not usable

rcgoodfellow:
Fixes
- #104 

With this PR we bail on the link reconciler if we cannot read an transceiver's MPN. According to SFF-8636, even for an unspecified Vendor Part Number (which is what we are calling the MPN), the transceiver must return all zeros. So failure to read this value means the transceiver is not (yet) functioning well enough for memory reads to work. If we proceed with setting up a link when the transceiver has not fully initialized, this leads to the issue @bnaecker notes [here](https://github.com/oxidecomputer/dendrite/issues/104#issuecomment-3330464690).

A precondition to reading the transceiver MPN is the lower level QSFP module initialization code determining that the transceiver is functioning well enough to be "supported". Here supported does not mean being a member of a qualified list, but rather passes a series of checks that give us confidence the device will work. So reading the MPN is really acting as a proxy for whether or not the device is likely to be usable.

### DIFF
--- a/dpd/src/link.rs
+++ b/dpd/src/link.rs
@@ -1653,7 +1653,7 @@ fn plumb_link(
     switch: &Switch,
     log: &slog::Logger,
     link: &mut Link,
-    mpn: &str,
+    mpn: &Option<String>,
 ) -> DpdResult<()> {
     debug!(log, "plumbing link");
     let connector = switch
@@ -1666,7 +1666,12 @@ fn plumb_link(
     // a default value for this transceiver.
     let fec = match link.config.fec {
         Some(fec) => Ok(fec),
-        None => switch.qsfp_default_fec(mpn),
+        None => match mpn {
+            None => Err(DpdError::Missing(
+                "Must specify FEC for unrecognized transceiver".to_string(),
+            )),
+            Some(mpn) => switch.qsfp_default_fec(mpn),
+        },
     }?;
 
     // Create the ASIC-layer's `TofinoPort`.
@@ -1729,7 +1734,7 @@ async fn reconcile_link(
     link_id: LinkId,
 ) {
     let mpn = {
-        let qsfp = match switch
+        let qsfp = switch
             .switch_ports
             .ports
             .get(&port_id)
@@ -1737,21 +1742,21 @@ async fn reconcile_link(
             .lock()
             .await
             .as_qsfp()
-            .cloned()
-        {
-            Some(qsfp) => qsfp,
-            None => {
-                warn!(log, "failed to get qsfp for port {port_id}");
-                return;
-            }
-        };
+            .cloned();
 
-        match qsfp_xcvr_mpn(&qsfp) {
-            Ok(mpn) => mpn,
-            Err(e) => {
-                warn!(log, "failed to get transceiver MPN {e}");
-                return;
+        if let Some(qsfp) = &qsfp {
+            match qsfp_xcvr_mpn(qsfp) {
+                Ok(mpn) => Some(mpn),
+                Err(e) => {
+                    warn!(
+                        log,
+                        "failed to get MPN for qsfp: {e}, port: {port_id}"
+                    );
+                    return;
+                }
             }
+        } else {
+            None
         }
     };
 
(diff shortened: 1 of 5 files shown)


rcgoodfellow:
A4x2 comes all the way up with this PR and the API works.

rcgoodfellow:
Packaging this into omicron, I'm able to bring berlin all the way up.

dpd/src/link.rs
```diff
@@ -1742,8 +1743,18 @@ async fn reconcile_link(
...
-        if let Some(qsfp) = qsfp {
-            qsfp_xcvr_mpn(&qsfp).unwrap_or(None)
+
+        if let Some(qsfp) = &qsfp {
+            match qsfp_xcvr_mpn(qsfp) {
+                Ok(mpn) => Some(mpn),
+                Err(e) => {
+                    warn!(
+                        log,
+                        "failed to get MPN for qsfp: {e}, port: {port_id}"
```
bnaecker, line 1753: I think we may want to use the key-value formatting in these log messages, something like:

```rust
warn!(log, "failed to get MPN for QSFP"; "port" => %port_id, "error" => %e);
```

Keeps them as distinct keys in the JSON output.

dpd/src/transceivers/mod.rs
```diff
@@ -236,7 +242,7 @@ mod mpn_test {
             transceiver: Some(transceiver),
             management_mode: ManagementMode::Manual,
         };
-        assert_eq!(qsfp_xcvr_mpn(&qsfp).unwrap(), None);
```
bnaecker, line 239: Can't comment on it, but the comment on L230 is out of date with this change.

dpd/src/types.rs
```diff
@@ -77,6 +79,9 @@ pub enum DpdError {
     Transceiver(#[from] TransceiverError),
     #[error("QSFP port \"{qsfp_port}\" has no transceiver")]
     MissingTransceiver { qsfp_port: QsfpPort },
+    /// Usability is currently determined by `check_module_support`.
+    #[error("The transceiver has failed basic operability checks")]
```
bnaecker, line 83: This is subjective, but I might be more specific about what "unusable" or "operability checks" mean. I would say "Could not identify the transceiver's management interface" or something pretty explicit like that.

dpd/src/types.rs
```diff
@@ -210,6 +215,12 @@ impl convert::From<DpdError> for dropshot::HttpError {
             DpdError::Faulted(e) => {
                 dropshot::HttpError::for_bad_request(None, e)
             }
+            DpdError::UnusableTransceiver => {
+                dropshot::HttpError::for_bad_request(
+                    None,
+                    "unusable transciever".to_string(),
```
bnaecker, line 221:
```suggestion
                    "unusable transceiver".to_string(),
```

dpd/src/types.rs
```diff
@@ -210,6 +215,12 @@ impl convert::From<DpdError> for dropshot::HttpError {
             DpdError::Faulted(e) => {
                 dropshot::HttpError::for_bad_request(None, e)
             }
+            DpdError::UnusableTransceiver => {
+                dropshot::HttpError::for_bad_request(
```
bnaecker, line 219: I have no idea if it matters to callers, but this doesn't seem like a 400 error. It seems like a 503 or something like that? One way it could matter is if we act on this to say, retry a request in the future, e.g., if we expect the vendor info to _become_ available shortly.

bnaecker (commented):
Thanks for tracking this down and fixing it! I've got a few small suggestions, but overall looks solid.

dpd/src/types.rs
rcgoodfellow: That information is not available to us by the time we report this error. All we have is a `Transceiver::Unsupported` enum at `transcievers/mod.rs:52. The reason I left the breadcrumb to `check_module_support` in the doc comment is so that it's easily discoverable what checks are in play and what other log messages to look for.

dpd/src/transceivers/mod.rs
rcgoodfellow: Fixed

PR #139, opened 2025-09-25
[xtask] migrate to clap v4

sunshowers:
I'm adding a new xtask, and it would be good to update to clap before I do that.


sunshowers:
might be obviated by https://github.com/oxidecomputer/dendrite/pull/134, depending on timing

PR #140, opened 2025-09-26
move asic-specific features into api trait

rcgoodfellow:
This moves remaining API methods into the API trait so the API manager can be used.


dpd-api/src/lib.rs
```diff
@@ -2075,3 +2256,218 @@ pub struct MulticastUnderlayGroupIpParam {
...
+/// The FEC counters for a specific link, including its link ID.
+#[derive(Clone, Debug, Deserialize, JsonSchema, Serialize)]
+pub struct LinkFecRSCounters {
+    /// The switch port ID.
+    pub port_id: PortId,
+    /// The link ID.
+    pub link_id: LinkId,
+    /// The FEC counter data.
+    pub counters: FecRSCounters,
+}
```
sunshowers, line 2280: You may want to put these in dpd-types, especially if they're going to be `replace`d in the future.

sunshowers (approved):
Thanks!

PR #160, opened 2025-11-10
chore: canonical exe path before validating.

jmcarp:
We validate that the dpd path lives at the expected location on start. However, this check can throw a false positive if we point to the exe via a symlink. This patch canonicalizes the exe path before validating so that we validate the real exe path, which may differ from env::current_exe().

Submitted to avoid a PATH workaround for running simulated omicron on macos. h/t @david-crespo.

### DIFF
--- a/asic/src/tofino_common/mod.rs
+++ b/asic/src/tofino_common/mod.rs
@@ -6,6 +6,7 @@
 
 use std::collections::HashMap;
 use std::ffi::OsStr;
+use std::fs;
 use std::path::PathBuf;
 
 use serde::Deserialize;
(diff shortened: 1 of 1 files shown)


izuzak:
:wave: Hi! I'm here since I ran into the same problem, reported [an issue on the omicron repo](https://github.com/oxidecomputer/omicron/issues/10367), and opened [an equivalent PR](https://github.com/oxidecomputer/dendrite/pull/264) before noticing that this PR existed. 🙈 

Definitely not urgent -- just wanted to give this PR a friendly bump since I hit the same problem and arrived at the same conclusion for a fix. :v:

cc @Nieuwejaar since I noticed you authored most of the recent commits in the repo history and maybe you can help get this over the 🏁 line.

Nieuwejaar (approved):

PR #167, opened 2025-12-15
Minor cleaning and repacking of p4 in preparation for external subnet support

Nieuwejaar:


zeeshanlakhani:
I'm guessing this is a part of something next, but this keeps the main logic intact (around mcast for sure).

PR #169, opened 2025-12-18
ipv6 fixes

rcgoodfellow:
- uplinkd: if addr is v6 create link local before static


rcgoodfellow:
closing in favor of #175

PR #177, opened 2026-01-05
External subnets

Nieuwejaar:
I've added both Ry and Levon as reviewers.  I'm particularly interested on Ry's thoughts on the P4 (which is implemented slightly differently on sidecar-lite) and Levon's input on the API, but I'll take any feedback you have to offer.

### DIFF
--- a/.github/buildomat/common.sh
+++ b/.github/buildomat/common.sh
@@ -1,9 +1,19 @@
 #!/bin/bash
 
+# The tofino2 has 20 stages, and the current sidecar.p4 needs all 20 of them.
+# Specifying the number of stages isn't strictly necessary, but it allows us to
+# track when we exceed the current ceiling. The underlying intention is to grow
+# deliberately and thoughtfully, given the limited space on the ASIC.
+#
+# Note: this now seems silly since we have maxed out the number of stages, but
+# we want to leave this check and note in place should we ever find a way to
+# reduce our footprint below 20 stages.
+TOFINO_STAGES=20
+
 # These describe which version of the SDE to download and where to find it
-SDE_COMMIT=eb08b0b0e55792bbba206202603da0fd791ea555
-SDE_PKG_SHA256=fbaea9ccd7ece70cdc20207e2f9b4750243468c6fe95c6a1fae9734201307db7
-SDE_DEB_SHA256=14050fd9a08b8b0b2ec7c827fa06a51516369a81bcf616d16699302b58d8ea9a
+SDE_COMMIT=e61fe02c3c1c384b2e212c90177fcea76a31fd4e
+SDE_PKG_SHA256=8a87a9b0bed3c5440a173a7a41361bdeb5e7a848882da6b4aa48c8fb0043f3bd
+SDE_DEB_SHA256=a292e2dd5311647c4852bb41f2532dd1fdf30325b6d04cccb7e85b873e521d5f
 
 [ `uname -s` == "SunOS" ] && SERIES=illumos
 [ `uname -s` == "SunOS" ] || SERIES=linux
--- a/common/src/network.rs
+++ b/common/src/network.rs
@@ -2,7 +2,7 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at https://mozilla.org/MPL/2.0/
 //
-// Copyright 2025 Oxide Computer Company
+// Copyright 2026 Oxide Computer Company
 
 use std::fmt;
 use std::net::Ipv6Addr;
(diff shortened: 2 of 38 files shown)


.github/buildomat/common.sh
```diff
@@ -1,5 +1,15 @@
...
 
+# The tofino2 has 20 stages, and the current sidecar.p4 needs all 20 of them.
+# Specifying the number of stages isn't strictly necessary, but it allows us to
+# track when we exceed the current ceiling. The underlying intention is to grow
+# deliberately and thoughtfully, given the limited space on the ASIC.
+#
+# Note: this now seems silly since we have maxed out the number of stages, but
+# we want to leave this check and note in place should we ever find a way to
+# reduce our footprint below 20 stages.
+TOFINO_STAGES=20
```
FelixMcFelix: Yikes 😬

common/src/network.rs
```diff
@@ -248,6 +248,125 @@ pub fn validate_vlan(id: impl Into<u16>) -> Result<(), VlanError> {
...
+// This is identical to the more generically named InstanceTarget, which should
+// be used for new APIs.
+#[derive(
+    Clone, Copy, Debug, PartialEq, Eq, Deserialize, JsonSchema, Serialize,
+)]
+pub struct NatTarget {
+    pub internal_ip: Ipv6Addr,
+    pub inner_mac: MacAddr,
+    pub vni: Vni,
+}
```
FelixMcFelix: Do we need to keep this type? I think replacing every use in the external API with `InstanceTarget` should be compatible as far as JSON serde is concerned.

(I could also be missing something with how our versioned Open API specs work.)

dpd-client/tests/integration_tests/external_subnet.rs
```diff
@@ -0,0 +1,685 @@
...
+    backplane_port: PhysPort, // switch port to which the gimlet is attached
+    backplane_ip: String,     // ip address of the gimlet's switch port
+
+    l4_protocol: L4Protocol,
+    geneve_vni: u32,
+}
+
+// On egress, we deencapsulate the geneve packet as we do for NAT, but we don't
+// rewrite the source address before forwarding the packet.
+async fn test_egress(switch: &Switch, test: &ExternalTest) -> TestResult {
```
FelixMcFelix: This comment confused me a bit. The first 'we' refers to the switch, but the second 'we' is the test harness I'd assume (since sidecar.p4 just decaps on egress)?

dpd-client/tests/integration_tests/external_subnet.rs
```diff
@@ -0,0 +1,685 @@
...
+        port: test.uplink_port,
+    };
+
+    switch.packet_test(vec![send], vec![expected])
+}
+
+// On ingress, we wrap the incoming packet with a geneve header and forward to
+// the correct gimlet.  Unlike NAT ingress, there is no manipulation of the
+// target address.
+async fn test_ingress(switch: &Switch, test: &ExternalTest) -> TestResult {
```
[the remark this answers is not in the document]
Nieuwejaar: Yeah, this is another comment on OPTE's NAT handling, which doesn't really have anything to do with this test.  I took it out.

dpd-client/tests/integration_tests/external_subnet.rs
```diff
@@ -0,0 +1,685 @@
...
+    switch: &Switch,
+    l4_protocol: L4Protocol,
+) -> TestResult {
+    let test = ExternalTest {
+        pkt_src_ip: "fa00:22::1".to_string(),
+        pkt_src_mac: "a1:a2:a3:a4:a5:a6".to_string(),
+        pkt_src_port: 3333,
+        pkt_dst_ip: "fc00:13::1".to_string(),
+        pkt_dst_mac: "b1:b2:b3:b4:b5:b6".to_string(),
+        pkt_dst_port: 4444,
```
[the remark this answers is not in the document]
Nieuwejaar: I think the src and dst IPs are correct, and match this diagram.  They are different in the ingress and egress cases.  That is, the endpoints' IP addresses don't change, but the src/dst polarity does.

I did have the external subnet wrong but, as you noted, that doesn't actually come into play in the egress case.

dpd/p4/sidecar.p4
```diff
@@ -135,6 +137,38 @@ control Filter(
 		ipv6_ctr.count();
 	}
 
+	action forward_extsub_v4_to(ipv6_addr_t target, mac_addr_t inner_mac, geneve_vni_t vni) {
```
rcgoodfellow: Along the same lines as [this comment](https://github.com/oxidecomputer/sidecar-lite/pull/129#discussion_r2663487004). It still feels weird to me that this is in the formerly named `Local` now named `Filter` control. Routing to an attached subnet feels a lot closer to `NatIngress` and I would have anticipated seeing a new control like `AttachedSubnetIngress`.

Issue #172, opened 2025-12-19
integration_tests::service::test_service_ipv4_unknown_address flake


rcgoodfellow:
I had `integration_tests::service::test_service_ipv4_unknown_address` fail in CI. Run is [here](https://buildomat.eng.oxide.computer/wg/0/details/01KCTDYNY8275QK96QFJ6HQW27/HRKNEcPo5YCPUzlsPQm1OiPFtWagRfdPNTXFBHTI02Qj2V5a/01KCTEPMG26R1PDWZ25GFRW7DG). Rerunning without changes was sufficient to have the test pass.

PR #181, opened 2026-01-08
add support for v4 routes over v6 nexthops

rcgoodfellow:
This PR adds support for adding IPv4 routes over IPv6 nexthops. This is needed for BGP unnumbered where nexthops are always IPv6 link local addresses.

There are no new tables added to this PR. Rather IPv6 variants of the exiting IPv4 actions on IPv4 router index table are added.

The API has a version bump to accommodate creating IPv4 routes over IPv6 nexhops as well is listing and fetching IPv4 routes.

This code has been tested in Maghemite CI where we have an `mgd` daemon feeding unnumbered routes to dendrite and they are installed successfully [here](https://github.com/oxidecomputer/maghemite/blob/0d429d28f7202d7ac7c2a2fc9179eefc7c5cfec7/falcon-lab/src/test.rs#L163-L174).

### DIFF
--- a/swadm/src/route.rs
+++ b/swadm/src/route.rs
@@ -239,9 +239,21 @@ async fn route_add(
             .await
             .context("adding IPv6 route")
             .map(|_| ()),
-        (IpAddr::V6(_), IpNet::V4(_)) => {
-            Err(anyhow!("cannot have an IPv4 route to an IPv6 address"))
-        }
+        (IpAddr::V6(tgt_ip), IpNet::V4(cidr)) => client
+            .route_ipv4_over_ipv6_add(&types::Ipv4OverIpv6RouteUpdate {
+                cidr,
+                target: types::Ipv6Route {
+                    tag: client.inner().tag.clone(),
+                    port_id: link_path.port_id,
+                    link_id: link_path.link_id,
+                    tgt_ip,
+                    vlan_id,
+                },
+                replace: false,
+            })
+            .await
+            .context("adding IPv4 or IPv6 route")
+            .map(|_| ()),
         (IpAddr::V4(_), IpNet::V6(_)) => {
             Err(anyhow!("cannot have an IPv6 route to an IPv4 address"))
         }


.github/buildomat/jobs/packet-test.sh
```diff
@@ -33,7 +35,7 @@ source .github/buildomat/linux.sh
 
 wd=`pwd`
 export WS=$wd
-STARTUP_TIMEOUT=${STARTUP_TIMEOUT:=15}
+STARTUP_TIMEOUT=${STARTUP_TIMEOUT:=45}
```
Nieuwejaar: I've never seen this be an issue before.  Are you seeing this in CI or somewhere else?

dpd-api/src/lib.rs
```diff
@@ -283,6 +284,48 @@ pub trait DpdApi {
     #[endpoint {
         method = GET,
         path = "/route/ipv4",
+        versions = ..VERSION_V4_OVER_V6_ROUTES
+    }]
+    async fn route_ipv4_list_v2(
```
Nieuwejaar: Why is this v2?  What was v1?

I've taken a very different approach to route versioning in the PR I sent you.

dpd-client/README-testing.md
```diff
@@ -71,7 +71,7 @@ All environment variables are prefixed by `DENDRITE_TEST_` for clarity.
   representation of packets on failure. The second bit controls whether to
   display the hex of each packet body as well.
 - `DENDRITE_TEST_TIMEOUT`: The amount of time to wait for any single test's
-  network traffic to complete, specified in milliseconds.  The default is 500,
+  network traffic to complete, specified in milliseconds.  The default is 1500,
```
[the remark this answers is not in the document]
Nieuwejaar: I suspect any timing issues you hit in CI were https://github.com/oxidecomputer/dendrite/issues/172.

rcgoodfellow:
Thanks @Nieuwejaar. I've added a packet test. That packet test surfaced some issues that I fixed.

This code will probably be on the default path for a4x2 as the datacenter topology will probably move to BGP unnumbered.

.github/buildomat/jobs/packet-test.sh
Nieuwejaar: If you're hitting this on a colo VM, then it seems reasonable to bump it up.

Nieuwejaar (approved):
Thanks.  This looks good.  I ticked the Approve box, but I would like it if you could get Rain to give the backwards compatibility stuff a quick look-over.

PR #189, opened 2026-01-22
[multicast] Narrow underlay scope to ff04::/64 and add ASM source filtering

zeeshanlakhani:
**Underlay changes**:
  - Restrict internal multicast from admin/site/org scoped (ff04, ff05, ff08) to just Omicron's reserved subnet (ff04::/64) as per lifecycle changes. These underlay addresses are made unique in Omicron.
  - Rename AdminScopedIpv6 to UnderlayMulticastIpv6.
  - Simplify P4 to only match ff04::/64.
  - Use omicron-common multicast constants for validation.
  - Update comments from "admin-scoped" to "admin-local" where appropriate

**Source filtering changes**:
  - Replace IpSrc::Subnet with IpSrc::Any for any-source matching.
  - Change IPv6 source filter from exact to LPM match.
  - Allow source filters on ASM groups (previously SSM-only).
  - Deduplicate sources in canonicalize_sources() via HashSet.

**API changes**:
  - API v7 (MCAST_SOURCE_FILTER_ANY): IpSrc::Any for ASM source filtering
  - API v8 (MCAST_STRICT_UNDERLAY): strict underlay subnet validation 
  - Add operation_id to versioned route and multicast endpoints that were missing them.
  - Add tag-based group ownership: delete and update endpoints *now* require a MulticastTag query parameter for group identity validation. V1 backward-compat handlers look up the existing tag automatically.

**P4/table changes**:
  - Skip external bitmap update for groups with no external members, avoiding spurious errors during rollback on empty groups.

### DIFF
--- a/dpd/src/types.rs
+++ b/dpd/src/types.rs
@@ -315,6 +315,7 @@ impl convert::From<common::network::VlanError> for DpdError {
     }
 }
 
+#[cfg(feature = "multicast")]
 impl convert::From<dpd_types::mcast::Error> for DpdError {
     fn from(err: dpd_types::mcast::Error) -> Self {
         DpdError::Invalid(err.to_string())


dpd/src/api_server.rs
```diff
@@ -2020,13 +2042,54 @@ impl DpdApi for DpdApiImpl {
...
+
+        // Lookup current tag for backward compat (v1-v5 clients may omit tag)
+        let tag = match entry.tag.as_ref() {
+            Some(t) => t.clone(),
+            None => {
+                mcast::get_group_internal(switch, underlay)
+                    .map_err(HttpError::from)?
+                    .tag
+            }
+        };
```
FelixMcFelix: Nit: this doesn't need to be a clone if you use a declaration without a definition, along the lines of

```suggestion
        // Lookup current tag for backward compat (v1-v5 clients may omit tag)
        let owned_tag;
        let tag = if let Some(t) = entry.tag.as_ref() {
            t
        } else {
            owned_tag = mcast::get_group_internal(switch, underlay)
                .map_err(HttpError::from)?
                .tag;
            &owned_tag
        };
```

dpd/p4/sidecar.p4
```diff
@@ -1785,19 +1785,20 @@ control MulticastIngress (
 	// Note: SSM tables currently take one extra stage in the pipeline (17->18).
 	apply {
 		if (hdr.geneve.isValid() && hdr.inner_ipv4.isValid()) {
-			// Check if the inner destination address is an IPv4 SSM multicast
-			// address.
-			if (hdr.inner_ipv4.dst_addr[31:24] == 8w0xe8) {
+			// Check if the inner destination address is an IPv4 multicast
+			// address (224.0.0.0/4). Apply source filtering for both SSM
+			// (232.0.0.0/8) and ASM ranges.
+			if (hdr.inner_ipv4.dst_addr[31:28] == 4w0xe) {
```
FelixMcFelix: Since the default action in these tables is to `drop`, does this change require that the control plane explicity passes `IpSrc::Any` when source filtering is not in play for ASM addresses?

PR #196, opened 2026-01-22
fix copyrights

Nieuwejaar:
This updates the copyrights on all the files changed during the big reformatting.


Nieuwejaar:
> @Nieuwejaar did you want to also just update the other files hhere, non-reformatted, in one go?

Nope.   I was just updating the files that were modified.

PR #203, opened 2026-01-28
[asic-details] Add lotnum (4-character lot number) to SwitchIdentifiers

### LINKED ISSUE #200: Want `swadm` command to print Tofino ASIC details
We currently collect a bunch of details about the Tofino ASICs in the managed Sidecar, including ASIC ID, fab, lot, wafer, wafer location, and more. We should add a Dendrite API and `swadm` command to print these details.

### LINKED ISSUE #202: Not reporting full lot numbers in `oximeter` statistics
It seems we're not reporting the full lot numbers correctly in our `oximeter` statistics about the switch (e.g, table occupancy). The real lot numbers are things like `TCAK77`, but we're only reporting values like `T`. It's not yet clear whether we're truncating that to the first character somewhere or what, but this needs to capture the right values.

zeeshanlakhani:
This PR adds lotnum (4-character lot number) to SwitchIdentifiers and exposes ASIC details via swadm (and the dpd-client API). The full lot identifier combines fab, lot, and lotnum (e.g., FTCAK7 instead of just T).

We do keep the current fields separated however, but update the view for the combination.

Adds v1::SwitchIdentifiers for API versions 1-4 (without lotnum), with the latest version (v5+) including the new field.

Fixes #200: Want swadm command to print Tofino ASIC details
Fixes #202: Not reporting full lot numbers in oximeter statistics


swadm/src/main.rs
```diff
@@ -91,6 +91,8 @@ enum Commands {
         #[command(subcommand)]
         cmd: compliance::Compliance,
     },
+    /// Display switch and ASIC identifiers.
```
Nieuwejaar: Maybe add 'id' as an alias?

zeeshanlakhani:
@rcgoodfellow I added the fuse info. I'll work on getting that piped through to oximeter when I update the schema in omicron next.

bnaecker:
I don't think we'll realistically be able to update the schema for switch stats right now. The update mechanism we built predates the online update system we have now. That means there's never a period when at least one switch isn't producing data, and so no window in which the old data and schema can be updated reliably to make way for the new.

What I would do here is put this more complete value into the existing `lot` field, for the time being. When we can get back to solving actual schema update here in the context of reconfigurator, then we can migrate the schema and data as needed.

PR #208, opened 2026-02-03
[sidecar/pipeline] P4 cleanup, getting us to 16 (non-mcast) / 19 (mcast) at current table sizes

zeeshanlakhani:
﻿﻿The sidecar pipeline lands at 16 ingress stages without multicast and 19 with multicast, all without sacrificing unicast LPM capacity in either build, closing https://github.com oxidecomputer/dendrite/issues/269.

#### P4/sidecar pipeline changes

  * Unify route_result_t struct for Router4/Router6 to prevent PHV liverange divergence
  * Unify nexthop: replace separate nexthop_ipv4/nexthop_ipv6 with single nexthop (ipv6_addr_t) plus nexthop_is_v6 flag
  * Add TTL compound key (idx, route_ttl_is_1) to route tables for TTL offload
    * This includes Rust updates and a larger routing table for the compound key
  * Add @pa_no_init pragmas for metadata fields to guard against compiler init bugs
  * Add @pa_container_type("normal") pragmas for all ingress booleans and wider metadata fields to prevent mocha container corruption (whole-container writes clobbering neighboring fields)
  * Extend @pa_container_type to egress bridge header fields and drop_reason
  * Zero sc_pad at every sc_code write site (was uninitialized before)
  * Bridge header now includes is_mcast_routed for CPU copy detection
  * v6 unicast TTL=1 handling via per-prefix skip_ttl bit + inline check (saves a stage vs the v4-style compound TTL key). Service-port routes skip the TTL exception so userspace still receives the packet

#### EgressFilter

  * Moves EgressFilter from ingress to egress pipeline to free one ingress stage for NatEgressFilter
  * Adds nat_egress_hit to bridge header to carry NAT state across TM boundary
  * Control Plane: updates table path and match key (egress_port instead of ucast_egress_port)

#### Egress MacRewrite

  * Creates separate unicast_mac_rewrite and mcast_mac_rewrite instances of MacRewrite
  * Unicast instance: rewrites src_mac only (dst_mac already correct from ARP/NDP)
  * Multicast instance: derives dst_mac from group address per RFC 1112/2464

#### Multicast Egress changes

  * Checks egress_rid != 0 first to identify PRE-replicated packets
  * Validates is_mcast_routed for CPU copies (egress_rid == 0 but routed to multicast)
  * Drops any CPU copies with DROP_MULTICAST_CPU_COPY reason

#### Counters

  * Moves Unicast, MulticastLL, EgressDropPort, EgressDropReason counters from multicast to base
  * Adds a Forwarded counter (for every packet copy that egresses the pipeline)
  * Removes Egress counter (from ingress pipe, replaced by per-port counters)
  * Removes MulticastDrop from MULTICAST_COUNTERS, as it's covered in general drop w/ reason
  * Handles link-local multicast counting in both MULTICAST and non-MULTICAST paths

#### Multicast table sizing

  * Per-workload constants in constants.p4 replace the single global ceiling. Ingress, router, replication, source filter, decap port and per-protocol mcast tables each sized independently.
  * v6 unicast LPM unified at 8192 across mcast and non-mcast builds (main defaulted to 1024 in mcast).

#### Build

  * Configures TOFINO_STAGES where there's 16 for base, 19 for multicast (the win!)


zeeshanlakhani:
Decided to target `main` as we're gated throughout for mcast's 18 stages.

zeeshanlakhani:
@taspelund + @FelixMcFelix, as promised, pinging you both here too.

zeeshanlakhani:
*Note*: If multicast is turned on, this will be needed because we're at 20 stages with unicast today.

zeeshanlakhani:
Closing, as we'll slowly bring this in piece-wise.

PR #215, opened 2026-02-06
api: unify IPv4 route endpoints to support IPv6 next-hops

taspelund:
The v4-over-v6 route feature lacked delete support entirely. Rather than add more separate endpoints, consolidate everything into /route/ipv4 under a new API version (v6).

Introduce Ipv4RouteUpdateV2 with RouteTarget (V4 or V6) for add/set, and widen route_ipv4_delete_target to accept IpAddr for the next-hop. Backward-compatible default impls keep the old /route/ipv4-over-ipv6 endpoints working in v4-v5.

Update swadm and integration tests to use the unified API.

### DIFF
--- a/openapi/dpd/dpd-6.0.0-eab964.json
+++ b/openapi/dpd/dpd-6.0.0-eab964.json
@@ -1,0 +1,9947 @@
+{
+  "openapi": "3.0.3",
+  "info": {
+    "title": "Oxide Switch Dataplane Controller",
+    "description": "API for managing the Oxide rack switch",
+    "contact": {
+      "url": "https://oxide.computer",
+      "email": "api@oxide.computer"
+    },
+    "version": "6.0.0"
+  },
+  "paths": {
+    "/all-settings": {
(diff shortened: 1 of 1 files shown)


dpd-api/src/lib.rs
```diff
@@ -2139,6 +2244,20 @@ pub struct Ipv4OverIpv6RouteUpdate {
     pub replace: bool,
 }
 
+/// Represents a new or replacement mapping of an IPv4 subnet to a single
+/// RouteTarget nexthop target, which may be either IPv4 or IPv6.
+#[derive(Debug, Clone, Deserialize, Serialize, JsonSchema)]
+pub struct Ipv4RouteUpdateV2 {
```
rcgoodfellow: can you name this and `RouteTargetIpv4PathV2` without the `V2` suffix and put them in a new file dpd-api/src/v2.rs similar to dpd-api/src/v1.rs

Nieuwejaar (approved):

PR #238, opened 2026-03-11
better tooling for PRBS mode for transceiver qual

### LINKED ISSUE #219: Want better tooling for PRBS mode for transceiver qual
Right now we have the ability to set ourselves into a PRBS mode; however, understanding that it's working is rather complicated and limited. The ucli has something like:

```
bf-sde.pm> prbs-show
prbs-show
Integration time: 10ms
...
-----------------------------------------------------------------------------------------------------------+--------------+----------+ ---------+---------+---------+---------+---------+---------+---------+-..
PORT |MAC |D_P|P/PT|PRBS  |    0     |    1     |    2     |    3     |    4     |    5     |    6     |    7     | Total Errors |   BER    |        1e-05     1e-06     1e-07     1e-08     1e-09     1e-10      1e-11
-----+----+---+----+----------+----------+----------+----------+----------+----------+----------+----------+--------------+----------+ ---------+---------+---------+---------+---------+---------+---------+-..
39/0 |18/0|272|2/16|PRBS7|         0|         0|         0|         0|          |          |          |          |            0 |  0.0e+00 | ---------+---------+---------+---------+---------+---------+---------+-..
40/0 |18/4|276|2/20|PRBS7|         0|         0|         0|         0|          |          |          |          |            0 |  0.0e+00 | ---------+---------+---------+---------+---------+---------+---------+-..
65/0 | 0/0|  2|0/ 2|PRBS7|  51674046|          |          |          |          |          |          |          |     51674046 |  5.0e-01 | 
```

There are a few things here that are worth calling out:

* Right now I had actually used `swadm link set-prop prbs 31` on 39/0 (qsfp6) and 40/0 (qsfp7), but it shows us in PRBS mode 7.
* It'd be good to get PRBS-specific counters out somehow when we're in this mode.

More generally, we're using PRBS modes as a way to try to understand the effectiveness of transceivers. When the gearbox doing FEC is in a dsp we lose the ease of the `link counters fec` logic that exists. The main things that we want to be able to get out is:

* How much PRBS data is being received?
* What are the observed error rates?

And use this as a proxy of link quality. It's hard to use the normal `rmon` counters because we're using PRBS rather than actual traffic.

Nieuwejaar:
Add support for monitoring PRBS error counts
Remove PRBS modes not supported by Tofino2
Reset PRBS state when deleting a link


dpd-api/src/lib.rs
```diff
@@ -2564,6 +2651,21 @@ pub trait DpdApi {
         path: Path<LinkPath>,
     ) -> Result<HttpResponseOk<Ber>, HttpError>;
 
+    /**
```
bnaecker: Style nit, but we should probably use the same comment delimiter here, `///`

dpd/src/link.rs
```diff
@@ -1281,6 +1285,28 @@ impl Switch {
         })?
     }
 
+    /// Return the measured error counts for a PRBS-enabled link over a
+    /// specified time period.
+    pub fn link_prbs_get_err(
+        &self,
+        port_id: PortId,
+        link_id: LinkId,
+        ms: u32,
```
bnaecker: Do we need to do any validation on this value of milliseconds? Someone could request both 0 and also `u32::MAX` here IIUC. That latter one especially seems bad.

bnaecker (commented):
This looks pretty good to me, thanks. I have a few suggestions, nothing major or structural.

bnaecker (approved):
Looks good, thanks for addressing my comments!

PR #243, opened 2026-03-31
put interrupt monitor behind a feature flag

Nieuwejaar:
There is still no root cause for #241, but it certainly seems likely that interrupts are involved.  The kernel code handling tofino/tfpkt interrupts hasn't changed in years, but `dpd` recently added code that monitors/manipulates interrupts in userspace.

Until the specific problem is identified, I would like to disable the userspace interrupt code with the hope that it will address the issue, allowing us to ship R19.  Unfortunately, since the appearance of the performance issue seems to be non-deterministic, I can't prove that this change will be effective even in the short term.  The best I can say is that I haven't seen a recurrence on madrid with these changes in place.


rcgoodfellow (approved):

rcgoodfellow:
The primary thing we lose here is the ability to detect TCAM parity errors. We should work toward being able to detect at least that, and if possible to enable/detect all RAS interrupts. But given the the issues we've seen in the lab that seem to be interrupt related, this seems prudent.

Nieuwejaar:
> The primary thing we lose here is the ability to detect TCAM parity errors. We should work toward being able to detect at least that, and if possible to enable/detect all RAS interrupts. But given the the issues we've seen in the lab that seem to be interrupt related, this seems prudent.

Agreed.  I'm going to keep digging for a root cause.

PR #258, opened 2026-04-14
Handle overflow reading near the end of a transceiver page

bnaecker:
- Promote u8s to u16s for math when computing the bounds of a large memory access, to fix overflow and reads right at the end of a page.
- Fixes #235


Nieuwejaar (approved):
Thanks.  I meant to go back and polish up the prototype fix, but it totally slipped my mind.

PR #259, opened 2026-04-19
[dpd-api] migrate to RFD 619 pattern

sunshowers:


dpd-api/src/lib.rs
```diff
@@ -596,13 +514,19 @@ pub trait DpdApi {
...
     ///
     /// This returns the unused MAC channels for each physical switch port. This can
     /// be used to determine how many additional links can be crated on a physical
     /// switch port.
+    //
+    // TODO: `FreeChannels` is unchanged across versions, so this split may be
+    // unnecessary — the rationale for gating the endpoint at
+    // `VERSION_MCAST_STRICT_UNDERLAY` is not obvious from the code alone.
+    // Revisit in a follow-up to either document the reason or collapse back to
+    // a single `channels_list` endpoint.
```
sunshowers: Wanted to ask about this -- is this correct?

dpd-api/src/lib.rs
```diff
@@ -1559,11 +1509,11 @@ pub trait DpdApi {
     #[endpoint {
         method = PUT,
         path = "/attached_subnet/{subnet}",
-	versions = VERSION_ATTACHED_SUBNETS..,
+        versions = VERSION_ATTACHED_SUBNETS..,
```
sunshowers, line 1488: There were a few tabs which I switched over to spaces.

dpd-api/src/lib.rs
zeeshanlakhani: We don't need the split. It was previously introduced because of a misspelling fix in the API doc comment, which shouldn't have warranted a version change (though it's official documentation). Collapsing it back to a single `channels_list` endpoint is totally the way to go.

dpd-api/src/lib.rs
```diff
@@ -2205,6 +2186,10 @@ pub trait DpdApi {
      * Underlay groups are used for admin-scoped IPv6 multicast traffic that
      * requires replication infrastructure with external and underlay members.
      */
+    // Required method: the latest version requires a `MulticastGroupTagQuery`
+    // query parameter that v1 does not have. When the tag is absent from the
+    // request body, the implementation must look up the existing group's tag
```
zeeshanlakhani: nit: v1's tag is always present in the body, but it's an Option vs "absent". What's missing is the concrete context type and its switch-specific ops. So maybe more like:

```suggestion
// Required method: v1 carries `tag` as `Option<String>` in the body, while the
// latest version moves it to a required `MulticastGroupTagQuery`. Resolving a
// `None` body tag means reading the existing group's tag from switch state,
// which is reachable only through the concrete context type in the implementation. 
```

dpd-api/src/lib.rs
```diff
@@ -2232,16 +2220,23 @@ pub trait DpdApi {
...
+    >;
 
     /**
      * Update an external-only multicast group configuration.
      *
      * Tags are optional for backward compatibility.
      */
+    // Required method: the latest version requires a `MulticastGroupTagQuery`
+    // query parameter that v7 does not have. When the tag is absent from the
+    // request body, the implementation must look up the existing group's tag
```
zeeshanlakhani: nit: follow a similar suggestion to above?

dpd-api/src/lib.rs
```diff
@@ -596,13 +514,19 @@ pub trait DpdApi {
...
     /// Get the set of available channels for all ports.
     ///
     /// This returns the unused MAC channels for each physical switch port. This can
     /// be used to determine how many additional links can be crated on a physical
     /// switch port.
+    //
+    // TODO: `FreeChannels` is unchanged across versions, so this split may be
+    // unnecessary — the rationale for gating the endpoint at
+    // `VERSION_MCAST_STRICT_UNDERLAY` is not obvious from the code alone.
+    // Revisit in a follow-up to either document the reason or collapse back to
```
zeeshanlakhani: Adding this again in context of review:

> We don't need the split. It was previously introduced because of a misspelling fix in the API doc comment, which shouldn't have warranted a version change (though it's official documentation). Collapsing it back to a single channels_list endpoint is totally the way to go (could do this quickly as a follow-up too).

zeeshanlakhani (approved):
Just a couple nits.

dpd-api/src/lib.rs
sunshowers: Ah interesting -- so we do actually require in case of doc comments to the latest version that a new version is added. So from that perspective it does make sense that a new version was added. I can try removing it now, though.

dpd-api/src/lib.rs
sunshowers: Thanks!

dpd-api/src/lib.rs
sunshowers: Done, thanks.

PR #264, opened 2026-05-04
Resolve potential symlink for dpd path when inferring P4 dir

izuzak:
See [this issue I opened on the omicron repo](https://github.com/oxidecomputer/omicron/issues/10367) for a more detailed description. 

TLDR version: `std::env::current_exe()` has [platform-specific behavior when the executable is invoked through a symlink](https://doc.rust-lang.org/std/env/fn.current_exe.html): some platforms return the symlink path, while others return the symlink target. Omicron exposes Dendrite through a symlinked bin directory, which caused `dpd` to infer the wrong install root on macOS (the symlink path). Using `std::fs::canonicalize()` removes that platform difference since, among other things, it [resolves symlinks](https://doc.rust-lang.org/std/fs/fn.canonicalize.html). With that, the path-shape validations will run on the symlink target instead of the symlink itself.

### DIFF
--- a/asic/src/tofino_common/mod.rs
+++ b/asic/src/tofino_common/mod.rs
@@ -348,6 +348,20 @@ fn infer_p4_dir() -> AsicResult<String> {
         AsicError::P4Missing(format!("looking up dpd path: {e:?}"))
     })?;
 
+    // std::env::current_exe() has platform-specific behavior when the
+    // executable is invoked through a symlink: some platforms return the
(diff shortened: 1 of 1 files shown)


izuzak:
Ahh, just as I opened this PR I noticed that @jmcarp opened an equivalent PR a few months ago: https://github.com/oxidecomputer/dendrite/pull/160. Closing this as a duplicate and will politely bump that other PR. :v:

PR #267, opened 2026-05-08
Ensure we use a stable DUID for DHCPv6 exchanges

bnaecker:
- Package `ndpd.conf` in the switch zone with defaults that preclude DHCPv6 on any interface.
- After fetching the correct, stable MAC addresses from the switch SP, `dpd` now uses the base MAC to write out a DUID to a file where illumos's `dhpcagent` can pick it up and use it later in exchanges.
- For the VLANs representing the techports, `tfportd` now creates both a link-local address and allows DHCPv6 to run on the interface as well. This should trigger `dhcpagent`, which would pick up the stable DUID from the previous item.
- Some misc cleanup, logging improvements, `IdOrdMap` over `BTreeMap`

### DIFF
--- a/.github/buildomat/illumos.sh
+++ b/.github/buildomat/illumos.sh
@@ -1,6 +1,10 @@
 # Download the SDE from CI, verify its integrity, and install it
 banner "sde setup"
 
+# Ensure the commands below don't abort this script if they succeed because
+# there was nothing to do.
+export PKG_SUCCESS_ON_NOP=1
+
 export PKG=tofino_sde.p5p
 curl -OL $SDE_DIR/$PKG
 SDE_CALC=`digest -a sha256 $PKG`
@@ -15,7 +19,7 @@ export LD_LIBRARY_PATH="$SDE/lib:$LD_LIBRARY_PATH"
 
 # Install a couple of non-standard packages needed to build dendrite
 banner "packages"
-pfexec pkg install clang-15 pcap
+pfexec pkg install clang-15 pcap gcc14
 pfexec pkg set-mediator -V 15 clang llvm
 
 cargo --version
--- a/dpd/src/dhcpv6/illumos.rs
+++ b/dpd/src/dhcpv6/illumos.rs
@@ -1,0 +1,210 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/
+//
+// Copyright 2026 Oxide Computer Company
+
+use anyhow::Context as _;
+use common::network::MacAddr;
+use scuffle::Scf;
+use slog::Logger;
+use slog::debug;
+use slog::error;
+use slog::info;
+use std::fmt::Write;
+use std::time::Duration;
+
+/// Path of the final resulting DUID file.
+const DUID_PATH: &str = "/etc/dhcp/duid";
+
+/// Path of the temp DUID file, so we can atomically swap it.
+const TEMP_DUID_PATH: &str = "/etc/dhcp/duid.temp";
+
+/// Interval on which we retry the various operations we need to succeed.
+const RETRY_INTERVAL: Duration = Duration::from_secs(5);
+
+/// List of techports we run DHCPv6 on.
+const TECHPORTS: [&str; 2] = ["techport0", "techport1"];
+
+/// Path of `in.ndpd`'s configuration file.
+const NDPD_CONF_FILE: &str = "/etc/inet/ndpd.conf";
+
+/// FMRI for the service running `in.ndpd
+const NDPD_FMRI: &str = "svc:/network/routing/ndp:default";
+
+/// Ensure that DHCPv6 is allowed on the techports.
+pub async fn allow_dhcpv6_on_techports(log: Logger, base_mac: MacAddr) {
+    // First, always ensure the DUID file is written correctly. We can do this
+    // with no coordination and it has to be done first.
+    ensure_duid_file_exists(&log, &base_mac).await;
+
+    // Now, rewrite the the NDP configuration file to allow DHCPv6 on the two
+    // techport interfaces.
+    rewrite_ndpd_conf(&log).await;
+
+    // Restart `in.ndpd`.
+    //
+    // This can happen before or after `tfportd` creates the link-local
+    // addresses using the new MAC. If the restart happens first, `in.ndpd` will
+    // start DHCPv6 if possible when the addresses are recreated. In the other
+    // order, `in.ndpd` will read all the addresses on startup and being NDP and
+    // DHCPv6 if possible.
+    restart_ndpd(&log).await;
+}
+
+/// Restart `in.ndpd`.
+///
+/// This blocks until the restart has occurred.
+async fn restart_ndpd(log: &Logger) {
+    loop {
+        let Err(e) = restart_ndpd_once().await else {
+            info!(log, "restarted `in.ndpd`");
+            return;
+        };
+        error!(
+            log,
+            "failed to start `in.ndpd`, will retry";
+            "error" => %e,
+        );
+        tokio::time::sleep(RETRY_INTERVAL).await;
+    }
+}
+
+async fn restart_ndpd_once() -> anyhow::Result<()> {
+    let scf = Scf::connect_global_zone()
+        .context("connecting to SCF in current zone")?;
+    let mut instance = scf
+        .instance_from_fmri(NDPD_FMRI)
+        .context("creating SMF instance from FMRI")?;
+    instance.smf_restart().context("restarting NDPD SMF service")
+}
+
+/// Write out new lines to the NDP configuration file allowing DHCPv6.
+///
+/// This blocks until the rewrite has occurred.
+async fn rewrite_ndpd_conf(log: &Logger) {
+    loop {
+        let Err(e) = rewrite_ndpd_conf_once().await else {
+            info!(
+                log,
+                "updated in.ndpd configuration file";
+                "path" => NDPD_CONF_FILE,
+            );
+            return;
+        };
+        error!(
+            log,
+            "failed to update in.ndpd configuration file, will retry";
+            "path" => NDPD_CONF_FILE,
+            "error" => %e,
+        );
+        tokio::time::sleep(RETRY_INTERVAL).await;
+    }
+}
+
+async fn rewrite_ndpd_conf_once() -> anyhow::Result<()> {
+    // NOTE: We completely replace the file.
+    //
+    // It might be safer to write only the parts we need, but it's very hard to
+    // ensure that we do that correctly, e.g., if there are partial writes. The
+    // cost is that this will be super confusing if we ever have more content in
+    // the `ndpd.conf` we ship with the switch zone. Those contents will be
+    // overwritten here.
+    let mut content = String::from("ifdefault StatefulAddrConf false\n");
+    for techport in TECHPORTS {
+        writeln!(&mut content, "if {techport} StatefulAddrConf true")
+            .with_context(|| {
+                format!("writing if line for techport {techport}")
+            })?;
+    }
+    tokio::fs::write(NDPD_CONF_FILE, content)
+        .await
+        .context("writing in.ndpd conf file")
+}
+
+/// Ensure our DHCPv6 Unique Identifier (DUID) is written persistently to disk.
+///
+/// Some deployments run DHCPv6 over our technician ports. In that protocol,
+/// clients identify themselves with a DUID, which is supposed to be a stable,
+/// unique identifier so that servers can assign consistent configuration data
+/// to the client. By default illumos's `dhcpagent` uses the Link-Layer Address
+/// Plus Time option for this ID. However both the time and MAC address can
+/// change, violating the stability requirement. The latter changes because the
+/// switch zone is assigned a VNIC as its "first" datalink from the sled-agent
+/// in the global zone. That VNIC has a random locally-administered prefix,
+/// 02:08:20:...
+///
+/// Once the `dhcpagent` has a DUID, it persists it to a file and reads that
+/// whenever starting a new exchange. In the Oxide product, we're ensuring this
+/// file contains our expected, stable ID, based on the MAC address stored in
+/// the switch's SP FRUID EEPROM.
+///
+/// This blocks until the file has been written.
+async fn ensure_duid_file_exists(log: &Logger, base_mac: &MacAddr) {
+    // Always overwrite the file. There's no harm to doing so, and it avoids
+    // potential races when we restart.
+    loop {
+        let Err(e) = write_duid_file_once(log, base_mac).await else {
+            info!(
+                log,
+                "wrote DUID based on MAC to disk";
+                "path" => DUID_PATH,
+                "MAC" => %base_mac,
+            );
+            return;
+        };
+        error!(
+            log,
+            "failed to write DUID to disk, will retry";
+            "path" => DUID_PATH,
+            "MAC" => %base_mac,
+            "error" => %e,
+        );
+        tokio::time::sleep(RETRY_INTERVAL).await;
+    }
+}
+
+/// Atomically write the DUID file once.
+async fn write_duid_file_once(
+    log: &Logger,
+    base_mac: &MacAddr,
+) -> anyhow::Result<()> {
+    // Write the DUID into a temporary file. Note that this needs to be on the
+    // same filesystem as the real one, or the rename will fail.
+    //
+    // This creates the file if needed, and replaces the entire contents if
+    // it already exists.
+    let bytes = super::create_duid_bytes(base_mac);
+    match tokio::fs::write(&TEMP_DUID_PATH, &bytes).await {
+        Ok(_) => debug!(log, "wrote DUID to tempfile"),
+        Err(e) => {
+            anyhow::bail!(
+                "failed to write DUID to tempfile '{}': {}",
+                TEMP_DUID_PATH,
+                e,
+            );
+        }
+    }
+
+    // Atomically swap it into place.
+    match tokio::fs::rename(&TEMP_DUID_PATH, DUID_PATH).await {
+        Ok(_) => {
+            info!(
+                log,
+                "wrote stable DHCPv6 DUID";
+                "path" => DUID_PATH,
+                "MAC" => base_mac.as_slice(),
+            );
+            Ok(())
+        }
+        Err(e) => {
+            anyhow::bail!(
+                "failed to rename DUID temp file '{}' to \
+                real path '{}': {}",
+                TEMP_DUID_PATH,
+                DUID_PATH,
+                e,
+            );
+        }
+    }
+}
--- a/tfportd/src/vlans.rs
+++ b/tfportd/src/vlans.rs
@@ -11,19 +11,27 @@ use std::net::Ipv6Addr;
 use anyhow::Context;
 use anyhow::anyhow;
 use anyhow::bail;
+use iddqd::IdOrdItem;
+use iddqd::IdOrdMap;
+use iddqd::id_upcast;
 use serde::Deserialize;
 use slog::error;
 use slog::info;
+use slog::warn;
 
 use crate::Global;
 use crate::linklocal;
 use crate::oxstats::link;
 use common::illumos;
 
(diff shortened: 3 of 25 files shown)


bnaecker:
This appears to work, barring the child-reaping issue discussed in https://github.com/oxidecomputer/tofino-sde/issues/15. But the actual functionality of writing the DUID file and starting DHCP on the techports works. I've confirmed that they start to send Solicit messages, and that they receive a response, albeit with no addresses because the lab environment has none. I'll add more testing details once my build without the reaping problem works its way through the pipeline.

I should note that we're _not_ doing the "usual" thing here of waiting for an NDP RA with the "Managed Configuration" bit set before starting DHCPv6 -- we do so unconditionally. We could in theory do that, by adding a line to `ndpd.conf` and then restarting `in.ndpd`. I chose this because it seemed less intrusive, but I'm open to the other route too.

.github/buildomat/common.sh
```diff
@@ -11,9 +11,9 @@
 TOFINO_STAGES=20
 
 # These describe which version of the SDE to download and where to find it
-SDE_COMMIT=2a6b33211c9675996dcb99fe939045506667ae94
-SDE_PKG_SHA256=d32739c368d1666b98dd74e25e22f83c209982e2c6670de6db5d6fdf49b5e275
-SDE_DEB_SHA256=3ecbf7c677bb722b351d5af74cee44fab70c1bb5eadc6ab2558ba714a8c3978b
+SDE_COMMIT=0af3643d410b7d06fce709b066287ce14e2e5b3d
```
rcgoodfellow: Just noting to move this once

- oxidecomputer/tofino-sde#16

lands

bnaecker:
I talked this through with @rmustacc earlier today. I think it's actually better to take the other path I mentioned above, restarting `in.ndpd` after reworking its configuration file to allow DHCPv6 on these interfaces. That does require a restart of that one daemon, but it also ensures that, after that restart, the rest of the system continues to work as usual. We don't need to continually ensure that DHCPv6 is running in `dpd`, for example, that's `in.ndpd`'s job.

I'm going to rework that bit of the code to reflect that now, and give this another test on a racklet.

dpd/src/dhcpv6/illumos.rs
```diff
@@ -0,0 +1,256 @@
...
+    const INTERVAL: Duration = Duration::from_secs(10);
+    loop {
+        info!(log, "starting DHCPv6 agent loop"; "base_mac" => %base_mac);
+        if let Err(e) = ensure_duid_file_exists(&log, &base_mac).await {
+            error!(
+                log,
+                "failed to ensure DUID file";
+                "error" => %e,
+            );
+            continue;
```
rcgoodfellow: some sort of `sleep` here to prevent a possible rage loop?

dpd/src/dhcpv6/illumos.rs
```diff
@@ -0,0 +1,256 @@
...
+        }
+    }
+}
+
+/// Check the error message from `ifconfig` to see if agent is already running.
+fn dhcp_is_already_running(e: &IllumosError) -> bool {
+    match e {
+        IllumosError::Exec(_) => false,
+        IllumosError::Failed(msg) => {
+            msg.contains(DHCP_IS_ALREADY_RUNNING_MSG)
```
rcgoodfellow: This feels pretty fragile but probably requires work outside dendrite to get to a better place. Let's create a follow up issue to sort out a more robust way of interacting with the system DHCP client.

tfportd/src/vlans.rs
```diff
@@ -165,39 +205,36 @@ pub async fn ensure_vlans(g: &Global, link: &str) -> anyhow::Result<()> {
...
-            match linklocal::create(g, name).await {
-                Ok(()) => {
-                    slog::debug!(g.log, "created link-local for vlan: {name}")
-                }
-                Err(e) => slog::error!(
-                    g.log,
-                    "failed to create link-local for vlan: {name}: {e}"
-                ),
-            }
+            let _ = linklocal::create(g, &info.name).await;
```
rcgoodfellow, line 237: Why remove the error (and debug) logging here?

rcgoodfellow (commented):
Thanks @bnaecker. Comments follow.

tfportd/src/vlans.rs
bnaecker: It was logged in the caller already. These functions are all a little confused about returning errors, logging them, expecting the caller to log etc. I was trying to normalize and clean up a bit while in this code.

dpd/src/dhcpv6/illumos.rs
bnaecker: This went away in 5a6ebf2. That restarts `in.ndpd` instead of using `ifconfig` manually.

dpd/src/dhcpv6/illumos.rs
bnaecker: Nice catch, but this is also gone in 5a6ebf2.

.github/buildomat/common.sh
bnaecker: Thanks! In 5a6ebf2, which restarts `in.ndpd` instead of calling `ipadm` or `ifconfig`, we don't actually need this change in the `tofino-sde` repo at all. I still think that's a good change, just independent of this PR if we go the new route.

dpd/src/dhcpv6/illumos.rs
```diff
@@ -0,0 +1,215 @@
...
+/// List of techports we run DHCPv6 on.
+const TECHPORTS: [&str; 2] = ["techport0", "techport1"];
+
+/// Path of `in.ndpd`'s configuration file.
+const NDPD_CONF_FILE: &str = "/etc/inet/ndpd.conf";
+
+/// FMRI for the service running `in.ndpd
+const NDPD_FMRI: &str = "svc:/network/routing/ndp:default";
+
+#[link(name = "scf")]
```
rcgoodfellow: possible to use https://github.com/illumos/libscf-sys here?

dpd/src/dhcpv6/illumos.rs
```diff
@@ -0,0 +1,215 @@
...
+        error!(
+            log,
+            "failed to start `in.ndpd`, will retry";
+            "error" => %e,
+        );
+        tokio::time::sleep(RETRY_INTERVAL).await;
+    }
+}
+
+async fn restart_ndpd_once() -> anyhow::Result<()> {
```
rcgoodfellow, line 73: Should this be called `refresh_ndpd_once` or was the underlying call meant to be `smf_restart_instance`? Feels like the verbs should line up.

dpd/src/dhcpv6/illumos.rs
bnaecker: That doesn't expose `smf_restart_instance()`, which is what I intended to use. AFAICT, `in.ndpd` doesn't handle a refresh, only a restart.

dpd/src/dhcpv6/illumos.rs
bnaecker: Nope, this was definitely supposed to be _restart_! I looked at the man page and typed the wrong thing anyway. Nice catch, thanks.

.github/buildomat/common.sh
bnaecker: I've reverted this for now. Since we're not spawning a child, there's no need to block this PR on that change.

dpd/src/dhcpv6/illumos.rs
rmustacc: @jgallagher has worked on https://github.com/oxidecomputer/scuffle which is a more rusty way to do this and does expose this I believe.

jgallagher: It doesn't expose `smf_restart_instance()` directly, but the functionality is there, yeah. It'd be something like this:

```rust
let scf = Scf::connect_global_zone()?;
let instance = scf.instance_from_fmri("your-instance-fmri")?;
instance.smf_refresh()?;
```

I need to get back to `scuffle` and actually publish it on crates.io, or you could pull it in as a git dependency for now if you'd like?

bnaecker: Thanks @rmustacc and @jgallagher. It looks like restarting the service is exposed through [this method](https://github.com/oxidecomputer/scuffle/blob/7181674535cc5ee0d461b2d53be19f181b9257e5/src/instance.rs#L648-L653), at least on Helios v3 bits. I think all the racklets are on that, as well as Omicron, so it should be safe. I'll give it a whirl in any case.

bnaecker: I've switched to scuffle in 3864764

bnaecker: After some fallout moving CI in this repo over to Helios v3, things are building again. I'm going to take this for one more lap on `berlin` with `scuffle` doing the SMF work.

bnaecker: I spoke too soon

bnaecker: Alright, with @citrus-it 's help, I've gotten past the Helios v2 -> v3 transition. I'm going for another spin on `berlin` shortly.

.github/buildomat/illumos.sh
```diff
@@ -19,7 +19,7 @@ export LD_LIBRARY_PATH="$SDE/lib:$LD_LIBRARY_PATH"
 
 # Install a couple of non-standard packages needed to build dendrite
 banner "packages"
-pfexec pkg install clang-15 pcap
+pfexec pkg install clang-15 pcap gcc-14
```
citrus-it: The package name for gcc is `gcc14`, without the dash.

citrus-it: Ah, I see the followup now.

bnaecker:
Alrighty, I've gone through one more round of rack-setup on `berlin`, and things look correct. Dendrite started up in 1986, and we see that it got the base MAC addresses from the Sidecar SP before time synced:

```
root@oxz_switch0:~# grep "base MAC" $(svcs -L dendrite) | looker
00:02:00.745Z INFO dpd: no base MAC address found, fetching from Sidecar FRUID
00:02:06.666Z INFO dpd: resetting base MAC address
    new = a8:40:25:05:35:02
    old = Temporary(a8:40:25:74:d6:f0)
```

Next, the DHCPv6 task that sets all this Rube Goldberg up starts. It then writes in the DUID to disk, updates the NDP configuration, and restarts `in.ndpd`:

```
root@oxz_switch0:~# grep dhcpv6-task $(svcs -L dendrite) | looker
00:02:06.667Z DEBG dpd: wrote DUID to tempfile
    unit = dhcpv6-task
00:02:06.668Z INFO dpd: wrote stable DHCPv6 DUID
    MAC = 0xA84025053502
    path = /etc/dhcp/duid
    unit = dhcpv6-task
00:02:06.668Z INFO dpd: wrote DUID based on MAC to disk
    MAC = a8:40:25:05:35:02
    path = /etc/dhcp/duid
    unit = dhcpv6-task
00:02:06.668Z INFO dpd: updated in.ndpd configuration file
    path = /etc/inet/ndpd.conf
    unit = dhcpv6-task
00:02:06.670Z INFO dpd: restarted `in.ndpd`
    unit = dhcpv6-task
```

The DUID and `ndpd.conf` files are correct:

```
root@oxz_switch0:~# cat /etc/inet/ndpd.conf
ifdefault StatefulAddrConf false
if techport0 StatefulAddrConf true
if techport1 StatefulAddrConf true
root@oxz_switch0:~# xxd /etc/dhcp/duid
00000000: 0003 0001 a840 2505 3502                 .....@%.5.
```

`in.ndpd` was started at the time indicated in the log file:

```
root@oxz_switch0:~# cat $(svcs -L ndp)
[ Dec 28 00:01:24 Disabled. ]
[ Dec 28 00:01:29 Enabled. ]
[ Dec 28 00:01:31 Executing start method ("/lib/svc/method/svc-ndp"). ]
[ Dec 28 00:01:35 Method "start" exited with status 0. ]
[ Dec 28 00:02:06 Stopping because service restarting. ]
[ Dec 28 00:02:06 Executing stop method (:kill). ]
[ Dec 28 00:02:06 Executing start method ("/lib/svc/method/svc-ndp"). ]
[ Dec 28 00:02:18 Method "start" exited with status 0. ]
```

And last, DHCPv6 is actually operating on these interfaces now:

```
root@oxz_switch0:~# ipadm | grep techport
techport0/ll      addrconf ok           fe80::aa40:25ff:fe05:3502%techport0/10
techport0/v6      static   ok           fdb1:a840:2504:211::1/10
techport0/ll      addrconf ok           fdb1:a840:2504:211:aa40:25ff:fe05:3502/64
techport1/ll      addrconf ok           fe80::aa40:25ff:fe05:3502%techport1/10
techport1/?       dhcp     ok           fd75:23e9:4173::f/128
techport1/v6      static   ok           fdb2:a840:2504:211::1/10
techport1/ll      addrconf ok           fdb2:a840:2504:211:aa40:25ff:fe05:3502/64
root@oxz_switch0:~# ifconfig techport0 inet6 dhcp status
Interface  State         Sent  Recv  Declined  Flags
techport0  SELECTING       10     0         0  [V6]
root@oxz_switch0:~# ifconfig techport1 inet6 dhcp status
Interface  State         Sent  Recv  Declined  Flags
techport1  BOUND            4     2         0  [V6]
(Began, Expires, Renew) = (05/20/2026 22:13, 05/21/2026 00:13, 05/20/2026 23:13)
```

The DHCPv6 server in the lab confirms it:

```
asilomar-arista-dmgmt#show dhcp server leases
fd75:23e9:4173::d
End: 2026/05/20 23:44:07 UTC
Last transaction: 2026/05/20 21:44:07 UTC
MAC address: a840.257d.0aea

fd75:23e9:4173::f
End: 2026/05/21 00:13:43 UTC
Last transaction: 2026/05/20 22:13:43 UTC
MAC address: a840.2505.3502

fd75:23e9:4173::ffff
End: 2026/05/21 00:14:30 UTC
Last transaction: 2026/05/20 22:14:30 UTC
MAC address: 1cfd.0878.c8de

asilomar-arista-dmgmt#show dhcp server ipv6
IPv6 DHCP server is active
DNS server(s):
DNS domain name:
Lease duration: 0 days 2 hours 0 minutes
Active leases: 3
IPv6 DHCP interface status:
   Interface     Status
---------------- -----------------
   Ethernet3     Active
   Ethernet33    Inactive (Not up)
   Ethernet34    Inactive (Not up)
   Ethernet47    Active
   Ethernet48    Active


Subnet: fd00:a:a:a::/127
Subnet name:
Range: fd00:a:a:a::1 to fd00:a:a:a::1 (0/1 addresses leased)
DNS server(s):
Direct: Inactive (No IPv6 addresses on interfaces match this subnet)
Relay: Active
Active leases: 0

Subnet: fd75:23e9:4173::/127
Subnet name:
Range: fd75:23e9:4173::1 to fd75:23e9:4173::1 (0/1 addresses leased)
DNS server(s):
Direct: Inactive (No IPv6 addresses on interfaces match this subnet)
Relay: Active
Active leases: 0

Subnet: fd75:23e9:4173::2/127
Subnet name:
Range: fd75:23e9:4173::3 to fd75:23e9:4173::3 (0/1 addresses leased)
DNS server(s):
Direct: Inactive (No IPv6 addresses on interfaces match this subnet)
Relay: Active
Active leases: 0

Subnet: fd75:23e9:4173::4/127
Subnet name:
Range: fd75:23e9:4173::5 to fd75:23e9:4173::5 (0/1 addresses leased)
DNS server(s):
Direct: Inactive (No IPv6 addresses on interfaces match this subnet)
Relay: Active
Active leases: 0

Subnet: fd75:23e9:4173::6/127
Subnet name:
Range: fd75:23e9:4173::7 to fd75:23e9:4173::7 (0/1 addresses leased)
DNS server(s):
Direct: Inactive (No IPv6 addresses on interfaces match this subnet)
Relay: Active
Active leases: 0

Subnet: fd75:23e9:4173::8/127
Subnet name:
Range: fd75:23e9:4173::9 to fd75:23e9:4173::9 (0/1 addresses leased)
DNS server(s):
Direct: Inactive (No IPv6 addresses on interfaces match this subnet)
Relay: Active
Active leases: 0

Subnet: fd75:23e9:4173::a/127
Subnet name:
Range: fd75:23e9:4173::b to fd75:23e9:4173::b (0/1 addresses leased)
DNS server(s):
Direct: Inactive (No IPv6 addresses on interfaces match this subnet)
Relay: Active
Active leases: 0

Subnet: fd75:23e9:4173::c/127
Subnet name:
Range: fd75:23e9:4173::d to fd75:23e9:4173::d (1/1 addresses leased)
DNS server(s):
Direct: Active
Relay: Active
Active leases: 1

Subnet: fd75:23e9:4173::e/127
Subnet name:
Range: fd75:23e9:4173::f to fd75:23e9:4173::f (1/1 addresses leased)
DNS server(s):
Direct: Active
Relay: Active
Active leases: 1

Subnet: fd75:23e9:4173::fffe/127
Subnet name:
Range: fd75:23e9:4173::ffff to fd75:23e9:4173::ffff (1/1 addresses leased)
DNS server(s):
Direct: Active
Relay: Active
Active leases: 1

```

It appears that the server is configured to only lease 1 address to this rack, because `techport0` is still attempting to actively select an address. I'm not sure how to confirm that on the switch. There are leasable ranges, but each only has 1 address in it. `techport1` looks like it got its address first, and presumably the server is configured so that `techport0` needs to be assigned out of the same range, which is now fully-leased.

@rcgoodfellow @Nieuwejaar and maybe @citrus-it or @jgallagher, let me know if y'all have any more feedback on this.

bnaecker:
Also, the last thing to confirm is that we're actually using the DUID-LL based on this stable MAC as the client identifier option in the Solicit message from the host:

```
^Croot@oxz_switch0:~# snoop -r -d techport0 -v dhcp6
Using device techport0 (promiscuous mode)
ETHER:  ----- Ether Header -----
ETHER:
ETHER:  Packet 1 arrived at 22:33:26.82491
ETHER:  Packet size = 122 bytes
ETHER:  Destination = 33:33:0:1:0:2, (multicast)
ETHER:  Source      = a8:40:25:5:35:2,
ETHER:  Ethertype = 86DD (IPv6)
ETHER:
IPv6:   ----- IPv6 Header -----
IPv6:
IPv6:   Version = 6
IPv6:   Traffic Class = 0
IPv6:   Flow label = 0x0
IPv6:   Payload length = 68
IPv6:   Next Header = 17 (UDP)
IPv6:   Hop Limit = 1
IPv6:   Source address = fe80::aa40:25ff:fe05:3502
IPv6:   Destination address = ff02::1:2
IPv6:
UDP:  ----- UDP Header -----
UDP:
UDP:  Source port = 546
UDP:  Destination port = 547 (DHCPv6S)
UDP:  Length = 68
UDP:  Checksum = 7535
UDP:
DHCPv6: ----- Dynamic Host Configuration Protocol Version 6 -----
DHCPv6:
DHCPv6: Message type (msg-type) = 1 (Solicit)
DHCPv6: Transaction ID = bf8004
DHCPv6:
DHCPv6: Option Code = 1 (Client Identifier)
DHCPv6:   DUID Type = 3 (Link-layer Address)
DHCPv6:   Hardware Type = 1 (Ethernet (10Mb))
DHCPv6:   Link Layer Address = a8:40:25:05:35:02
DHCPv6: Option Code = 3 (Identity Association for Non-temporary Addresses)
DHCPv6:   IAID = 80
DHCPv6:   T1 (renew) = 0 seconds
DHCPv6:   T2 (rebind) = 0 seconds
DHCPv6: Option Code = 6 (Option Request)
DHCPv6:   Requested Option Code = 7 (Preference)
DHCPv6:   Requested Option Code = 12 (Server Unicast)
DHCPv6:   Requested Option Code = 23 (DNS Recursive Name Server)
DHCPv6:   Requested Option Code = 24 (Domain Search List)
DHCPv6:   Requested Option Code = 27 (Network Information Service Servers)
DHCPv6:   Requested Option Code = 29 (Network Information Service Domain Name)
DHCPv6: Option Code = 14 (Rapid Commit)
DHCPv6: Option Code = 8 (Elapsed Time)
DHCPv6:   Elapsed Time = 655.35 seconds
DHCPv6:

```

rcgoodfellow (approved):

bnaecker:
Alright, after some last-minute help from Ry, I've convinced myself this is in fact working as intended. We get DHCPv6 addresses on `berlin`'s `techport1` on each of the two switches (not `techport0`). Those are the ports connected to the Arista switch in the lab, running the DHCPv6 server.

PR #272, opened 2026-05-15
Allow in place update of physical link settings

internet-diglett:
Enable "in place" updates to existing link settings


jgallagher:
Tested this against with the `tofino_stub` dpd against the property-based tests in https://github.com/oxidecomputer/omicron/pull/10449, and they're happy with this change.

Levon and I started to test this on a racklette before realizing we don't have any supported transceivers that can change speeds; a real test would require physically swapping transceivers before reconfiguring via Nexus. Our thought was that that's still a thing we should do at some point, but it doesn't seem particularly urgent.

Nieuwejaar (approved):

PR #273, opened 2026-05-15
Implement shrink-in-place logic for LPM next-hop deletions

taspelund:
This PR implements shrink-in-place logic for LPM next-hop (`route_target`) deletions (i.e. the `replace_route_targets()` codepath). Shrink-in-place logic avoids a fallible allocation of new next-hop entries, which can be observed when attempting to delete an ECMP next-hop when the table is genuinely full or too fragmented for a contiguous allocation to succeed.

The PR introduces test cases for next-hop deletion both when the table is full and when the table is fragmented.

Additionally, the PR refactors some of the error handling paths in replace_route_targets, adds a test harness for simulating ASIC operation failures, and introduces test cases for the error/fallback paths.


taspelund:
Link to customer issue: https://github.com/oxidecomputer/customer-support/issues/1141

dpd/src/main.rs
```diff
@@ -258,7 +258,7 @@ fn get_sidecar_revision(
 }
 
 impl Switch {
-    fn new(
+    pub(crate) fn new(
```
Nieuwejaar: Why is this needed?  Is this a holdover from before you added the RouteTableOps trait?

dpd/src/route.rs
```diff
@@ -293,6 +295,126 @@ impl RouteData {
...
+/// invoking the ASIC.  Production code passes `&Switch`, whose impl below
+/// dispatches to the live `table::route_ipv{4,6}` functions.
+trait RouteTableOps {
+    /// Logger to use for the routing subsystem's messages.
+    fn log(&self) -> &slog::Logger;
+
+    /// Size (in slots) of the `route_target` forwarding table for the
+    /// requested family.  Used by `add_route_locked` to lazily initialize
+    /// the per-family `FreeMap` on first use.
+    fn route_fwd_table_size(&self, is_ipv4: bool) -> DpdResult<u32>;
```
Nieuwejaar: It seems weird that this is a switch operation.  Can't you just look at `table.usage.size`?

dpd/src/route.rs
```diff
@@ -293,6 +295,126 @@ impl RouteData {
...
+
+    /// Write `target` into slot `idx` of the `route_target` table for a
+    /// subnet of the indicated family.  Production dispatches on
+    /// `(subnet_is_ipv4, target.route.tgt_ip)`: an IPv4 target always goes
+    /// to the v4 table; a v6 target goes to `route_ipv4::add_route_target_v6`
+    /// when the subnet is v4, otherwise to the v6 table.
+    fn add_target(
+        &self,
+        subnet_is_ipv4: bool,
```
Nieuwejaar: Why not just pass in the subnet?  I know we would only use it to determine whether it's v4 or v6, but it seems more natural.  Right now half the operations take the subnet and half take the bool.  It seems like it would be cleaner if they all took the subnet.

dpd/src/route.rs
```diff
@@ -448,86 +650,280 @@ fn replace_route_targets(
...
+    // slot's ASIC entry is currently deleted.
+    let mut live: Vec<Option<NextHop>> =
+        old.targets.iter().map(|t| Some(t.clone())).collect();
+    // Positions whose ASIC contents have diverged from `old.targets[i]`.  A
+    // position lands here as soon as we successfully delete its slot; from
+    // that point forward it differs from the original, regardless of whether
+    // the subsequent write succeeded.  Rollback walks this set instead of
+    // all `old.slots` positions, which keeps recovery O(removed) and lets us
+    // log per-position failures without flooding for untouched survivors.
+    let mut touched: Vec<u16> = Vec::with_capacity(removed.len());
```
Nieuwejaar: Do you really need this?  It seems like you could just iterate over `live` and then rewrite any entry that doesn't match `old.targets`.  The comment suggests that this is intended to make rollback more efficient.  Given that we expect rollback to be vanishingly infrequent, and `n` isn't that large, I wouldn't worry too much about efficiency.

Nieuwejaar (commented):
I still need to review all the new tests.

The new stuff looks good, and the extensive commenting is great.  The changes I've suggested are more stylistic than functional.

taspelund:
feedback addressed in latest commit

rcgoodfellow:
Can we add an integration level test for this? E.g. in dpd-client/tests/integration_tests. That way we have test coverage against the stack running against a simulated tofino.

taspelund:
Ok, integration and unit tests failed with the temporary revert of the fix -- just like we expected and wanted to happen.
Just yanked out the revert. Will confirm the integration and unit tests now pass.

rcgoodfellow (approved):

sion42x:
Using https://github.com/oxidecomputer/omicron/tree/rel/v21/rc0 to reproduce the issue on Madrid:

dpd builds:

- `oxz_switch1`: pre-fix, deployed v21 release binary
- `oxz_switch0`: post-fix, built from PR #273

### Method

- 3-way ECMP built with: `swadm route add <cidr> <port/link> <gw>` appends a next-hop; 3 `(port,gw)` per prefix gives a 3-slot target run.
- Fill `100.128.0.0/10` /24s, 3-way each, until `swadm route add` stops installing
- Shrink: `swadm route del <cidr> <qsfp2/0> <gw>` for every prefix (remove one next-hop). Exercises `replace_route_targets` 3-to-2 on a full table.
- Measured: per-prefix ECMP width before and after (`swadm route get`), alloc-error log lines during the shrink, 507 count by request method.

### Result

| switch | dpd | full at | width before | width after | shrink alloc-error lines |
|---|---|---|---|---|---|
| oxz_switch1 | pre-fix | 2043 | 3 | 3 | 4084 |
| oxz_switch0 | post-fix | 1699 | 3 | 2 | 10 |

- pre-fix: shrink `alloc(2)` fails on the full table; `finalize_route` restores the old 3-way entry; target not removed; `del` returns 507. 4084 is ~2 lines per failed del over 2043 prefixes.
- post-fix: shrink-in-place, no allocation; every prefix goes to 2-way. The 10 error lines were `POST /route/ipv4` add-rejections (probably mgd/reconciler against the still-full table) during the 6.5 min shrink window. 507-by-method in that window: 0 DELETE, 1 POST. Shrink failures: 0.

### Evidence (raw script output)

pre-fix (oxz_switch1):

```
POST-SHRINK (sw1)
    100.128.0.0/24 width 3->3    100.131.253.0/24 width 3->3
    alloc/Route2Index errors during shrink: 4084
```

post-fix (oxz_switch0):
```
POST-SHRINK (sw0)
    100.128.0.0/24 width 3->2    100.129.168.0/24 width 3->2    100.131.81.0/24 width 3->2
    alloc/Route2Index errors during shrink: 10
507 by method in window: 0 DELETE, 1 POST
```

pre-fix del on a full table:
```
Error: deleting IPv4 route target
    status: 507 Insufficient Storage; value: Error { error_code: None, message: "Route2Index" }
```

### Notes (worth investigation):

1. Full-table wedge (pre-fix). On a full table `route del` of a next-hop also 507s (same realloc). The table cannot shrink to recover or accept adds until `svcadm restart dendrite`. Observed: CLEAR of 2043 prefixes removed 0; count stayed 2042.

2. Possible route_ipv4 freemap slot leak. Freemap reports full well below capacity: oxz_switch0 returned `507 "Route2Index"` on a single add at 1045 routes (~3135 of ~8192 slots). Restart-recoverable. Undetermined whether shrink-in-place fails to return freed slots.

PR #321, opened 2026-07-14
Fix FreeMap leak during recycle bin reclaim

taspelund:
During reclaim where we try to consolidate contiguous free slots, the entirety of the tracked free slots was overwritten by the recycle bins being consolidated. This resulted in all free items being leaked, as we lose all references to them when they're overwritten.

This PR modifies the reclaim operation to include the free list with the recycle bins when attempting consolidation, then assigning the combination to the free list. This ensures entries are not lost.
i.e.
Before: consolidate recycle bins, assign results to free list
After: consolidate recycle bins AND free list contents, assign results to free list

### DIFF
--- a/dpd/src/freemap.rs
+++ b/dpd/src/freemap.rs
@@ -370,3 +370,21 @@ fn test_reclaim() -> anyhow::Result<()> {
 
     Ok(())
 }
+
+#[test]
+fn test_reclaim_preserves_existing_freelist() -> anyhow::Result<()> {
+    let mut map = new_freemap(8);
+
+    // Leave a three-slot span in the freelist, allocate its first slot, then
+    // return that slot to the recycle bin.  The physically contiguous free
+    // range [5, 8) is now split between the recycle bin and freelist.
+    assert_eq!(map.alloc(5u16)?, 0);
+    let recycled = map.alloc(1u16)?;
+    assert_eq!(recycled, 5);
+    map.free(recycled, 1u16);
+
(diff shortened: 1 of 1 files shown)


sion42x (approved):
From tests on madrid:

route_index (lookup): 2740 entries, width distribution {3-way: 1024, 2-way: 1697, 1-way: 19}

1024 3-way: the BGP-announced 100.64 fill, not touched by the shrink
1697 2-way + 18 1-way: the swadm 100.128 fill, 3-to-2 shrunk (18 tail entries only reached 1-way during the fill as the table filled)

Slot accounting:

* sum of claimed slots = distinct claimed slots = route_data used = 6485
* overlapping indices: 0
* route_data slots not referenced by any lookup entry (orphaned/leaked): 0
* lookup entries whose slot range is not fully present in route_data (dangling): 0

Free space:

1707 free slots across 1704 runs: 1703 of length 1, one of length 4
runs of length >=2: 1; runs of length >=3: 1

All of which is to say: looks good, still fragmentation obviously but all slots are present and accounted for as expected with this fix.

dpd/src/freemap.rs
rcgoodfellow, line 217: So the leak took place here before, basically removing pre-existing entries from the free list and only retaining newly recycled ones. Is that right? Can you add a brief description of the algorithmic change to the PR description?

taspelund: Done

rcgoodfellow (approved):

PR #330, opened 2026-07-20
Ensure provided port IDs are valid ASCII

bnaecker:


bnaecker:
ping @rcgoodfellow

common/src/ports.rs
```diff
@@ -92,7 +92,13 @@ macro_rules! make_port_type {
                 if s.len() <= $prefix.len() {
                     return Err("Invalid port kind");
                 }
```
FelixMcFelix: We can remove this now that we're using `split_at_checked` below.

FelixMcFelix (approved):
Saw this while working through my inbox -- this looks good, just one nit.

common/src/ports.rs
bnaecker: Fixed in 0570fc5

PR #334, opened 2026-08-06
[1/3] [nat] move `PortRange` to struct

nicolaskagami:
This PR:
- Moves the `PortRange` trait into a struct, validating the `low <= high` invariant by construction.
- Simplifies `overlap` function.

This is the first of 3 PRs simplifying and de-duplicating some of the `nat.rs` code.
- #334
- #335
- #336

Left for future work:
- Expanding `PortRange` usage further.
- Storing ranges in a way that codifies its non-overlapping invariant and stores sorted for faster operations.

### DIFF
--- a/dpd/src/nat.rs
+++ b/dpd/src/nat.rs
@@ -16,52 +16,64 @@ use crate::types::{DpdError, DpdResult};
 use common::nat::{Ipv4Nat, Ipv6Nat};
 use common::network::NatTarget;
 
-trait PortRange {
-    fn low(&self) -> u16;
-    fn high(&self) -> u16;
+/// An inclusive range of ports, guaranteed by construction to have
+/// `low <= high`.
+#[derive(Clone, Copy, PartialEq)]
+pub(crate) struct PortRange {
+    low: u16,
+    high: u16,
+}
+
+#[derive(Debug)]
+pub(crate) struct InvalidPortRange;
+
+impl From<InvalidPortRange> for DpdError {
+    fn from(_: InvalidPortRange) -> Self {
+        DpdError::Invalid("invalid port range".into())
+    }
+}
+
+impl PortRange {
+    fn new(low: u16, high: u16) -> Result<Self, InvalidPortRange> {
+        if low <= high {
+            Ok(PortRange { low, high })
+        } else {
+            Err(InvalidPortRange)
+        }
+    }
+
+    fn overlaps(self, other: PortRange) -> bool {
+        self.low <= other.high && self.high >= other.low
+    }
+}
+
+impl fmt::Display for PortRange {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "[{}-{}]", self.low, self.high)
+    }
 }
 
 #[derive(PartialEq)]
 pub(crate) struct Ipv6NatEntry {
-    pub low: u16,
-    pub high: u16,
+    pub l4_ports: PortRange,
     pub tgt: NatTarget,
-}
-
-impl PortRange for Ipv6NatEntry {
-    fn low(&self) -> u16 {
-        self.low
-    }
-    fn high(&self) -> u16 {
-        self.high
-    }
 }
 
 impl fmt::Display for Ipv6NatEntry {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        write!(f, "[{}-{}] -> {}", self.low, self.high, self.tgt)
+        write!(f, "{} -> {}", self.l4_ports, self.tgt)
     }
 }
 
 #[derive(Clone, PartialEq)]
 pub(crate) struct Ipv4NatEntry {
-    pub low: u16,
-    pub high: u16,
+    pub l4_ports: PortRange,
     pub tgt: NatTarget,
-}
-
-impl PortRange for Ipv4NatEntry {
-    fn low(&self) -> u16 {
-        self.low
-    }
-    fn high(&self) -> u16 {
-        self.high
-    }
 }
 
 impl fmt::Display for Ipv4NatEntry {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        write!(f, "[{}-{}] -> {}", self.low, self.high, self.tgt)
+        write!(f, "{} -> {}", self.l4_ports, self.tgt)
     }
 }
 pub struct NatData {
@@ -78,51 +90,38 @@ fn ipv4_entry(ipv4: Ipv4Addr, e: &Ipv4NatEntry) -> String {
     format!("{ipv4}/{e}")
 }
 
-fn overlaps<T: PortRange>(e: &T, low: u16, high: u16) -> bool {
-    let elow = e.low();
-    let ehigh = e.high();
-
-    (elow >= low && elow <= high)
-        || (ehigh >= low && ehigh <= high)
-        || (elow <= low && ehigh >= high)
-}
-
 /// find index of first mapping that overlaps with supplied port range
-fn find_first_mapping<T: PortRange>(
-    entries: &[T],
-    low: u16,
-    high: u16,
+fn find_first_mapping(
+    mut ranges: impl Iterator<Item = PortRange>,
+    range_to_find: PortRange,
 ) -> Option<usize> {
-    entries.iter().position(|e| overlaps(e, low, high))
+    ranges.position(|e| e.overlaps(range_to_find))
 }
 
 /// find indices of all mappings that overlap with supplied port range
-fn find_mappings<T: PortRange>(
-    entries: &[T],
-    low: u16,
-    high: u16,
+fn find_mappings(
+    ranges: impl Iterator<Item = PortRange>,
+    range_to_find: PortRange,
 ) -> Vec<usize> {
-    entries
-        .iter()
+    ranges
         .enumerate()
-        .filter(|(_, e)| overlaps(*e, low, high))
+        .filter(|(_, e)| e.overlaps(range_to_find))
         .map(|(i, _)| i)
         .collect()
 }
 
-fn find_space<T: PortRange>(
-    entries: &[T],
-    low: u16,
-    high: u16,
+fn find_space(
+    ranges: impl ExactSizeIterator<Item = PortRange>,
+    candidate_range: PortRange,
 ) -> Option<usize> {
-    let len = entries.len();
+    let len = ranges.len();
+    let iter = ranges.enumerate();
 
-    for (idx, e) in entries.iter().enumerate() {
-        if overlaps(e, low, high) {
+    for (idx, e) in iter {
+        if e.overlaps(candidate_range) {
             return None;
         }
-        if e.low() >= high && (idx == len - 1 || entries[idx + 1].low() >= high)
-        {
+        if e.low >= candidate_range.high {
             return Some(idx);
         }
     }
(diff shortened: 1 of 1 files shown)


dpd/src/nat.rs
```diff
@@ -16,52 +16,64 @@ use crate::types::{DpdError, DpdResult};
...
     }
 }
 
-#[derive(Clone, PartialEq)]
-pub(crate) struct Ipv4NatEntry {
-    pub low: u16,
-    pub high: u16,
+#[derive(PartialEq)]
+pub(crate) struct Ipv6NatEntry {
+    pub ports: PortRange,
```
internet-diglett: minor nit: `port` is a very overloaded term inside of dendrite (grepping for `ports` gives a lot of unrelated results). Maybe we should be more explicit and name this field `port_range`? I also notice that there is a convention of naming fields that are a single item `thing` and naming collections / iterators of a type `thing**s**`, so this also avoids creating confusion there.

dpd/src/nat.rs
```diff
@@ -16,52 +16,64 @@ use crate::types::{DpdError, DpdResult};
...
-        self.high
+impl fmt::Display for Ipv6NatEntry {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{} -> {}", self.ports, self.tgt)
     }
 }
 
+#[derive(Clone, PartialEq)]
+pub(crate) struct Ipv4NatEntry {
+    pub ports: PortRange,
```
internet-diglett: same as above

dpd/src/nat.rs
```diff
@@ -78,50 +90,39 @@ fn ipv4_entry(ipv4: Ipv4Addr, e: &Ipv4NatEntry) -> String {
...
 
-    for (idx, e) in entries.iter().enumerate() {
-        if overlaps(e, low, high) {
+    while let Some((idx, e)) = iter.next() {
+        if e.overlaps(range) {
             return None;
         }
-        if e.low() >= high && (idx == len - 1 || entries[idx + 1].low() >= high)
+        if e.low >= range.high
+            && iter.peek().is_none_or(|(_, next)| next.low >= range.high)
```
internet-diglett:
```rust
&& iter.peek().is_none_or(|(_, next)| next.low >= range.high)
```

🤔 Are we sure this check is still necessary? If so, can we provide commentary around what particular edge case we're protecting against here?

dpd/src/nat.rs
```diff
@@ -131,41 +132,42 @@ fn find_space<T: PortRange>(
...
+    assert_eq!(first_mapping(19, 40), None);
+    assert_eq!(first_mapping(0, 0), None);
+    assert_eq!(first_mapping(0, 2), Some(0));
+    assert_eq!(space(0, 0), Some(0));
+    assert_eq!(space(0, 1), None);
+    assert_eq!(space(11, 11), Some(2));
+    assert_eq!(space(19, 32), Some(3));
+    assert_eq!(space(0, 2), None);
+    assert_eq!(space(3, 5), None);
+    assert_eq!(space(3, 8), None);
```
internet-diglett, line 169: Could this be a good candidate for property based testing?

internet-diglett (commented):
Overall looks good! Just a few comments.

dpd/src/nat.rs
nicolaskagami: Nice catch!

dpd/src/nat.rs
nicolaskagami: I guess so, but in this case it would be enforcing a property that's basically the same as what we already have in the code. I like to think of tests like this as a way of giving the developer a feel for how it works more so than actually exercising functionality.

internet-diglett: I guess that was a roundabout way of me asking "it seems we need a lot of cases to exercise this code... are we sure we've covered all of them?"

(I'm actually not sure either way, it just stood out to me)

dpd/src/nat.rs
bnaecker: I think we use "L4 port" in Omicron and OPTE similar situations, which might be a good option.

dpd/src/nat.rs
nicolaskagami: It's checking that the next range either doesn't exist or starts after the one we're trying to place. It's still quite necessary.

dpd/src/nat.rs
nicolaskagami: Are you suggesting `PortRange` be called `L4PortRange`?

dpd/src/nat.rs
nicolaskagami: I just adapted the tests that were already there since this purposefully a quite mechanical PR with no new semantics.

dpd/src/nat.rs
bnaecker: Yeah, sorry if that wasn't clear. One could also rename the `ports` field to something like `l4_ports`.

nicolaskagami: Just making sure. Sounds good to me :+1:

dpd/src/nat.rs
```diff
@@ -78,50 +90,39 @@ fn ipv4_entry(ipv4: Ipv4Addr, e: &Ipv4NatEntry) -> String {
...
-}
-
 /// find index of first mapping that overlaps with supplied port range
-fn find_first_mapping<T: PortRange>(
-    entries: &[T],
-    low: u16,
-    high: u16,
+fn find_first_mapping(
+    mut ranges: impl Iterator<Item = PortRange>,
+    range: PortRange,
```
taspelund: I find the arg names here to be a bit confusing because they're so similar.
If I'm looking only at the body of this function, I don't know what the difference is between `ranges` and `range` is. Perhaps renaming `range` to something like search target would be helpful in making it immediately obvious what the semantics are for the arg.

dpd/src/nat.rs
```diff
@@ -78,50 +90,39 @@ fn ipv4_entry(ipv4: Ipv4Addr, e: &Ipv4NatEntry) -> String {
...
 }
 
 /// find indices of all mappings that overlap with supplied port range
-fn find_mappings<T: PortRange>(
-    entries: &[T],
-    low: u16,
-    high: u16,
+fn find_mappings(
+    ranges: impl Iterator<Item = PortRange>,
+    range: PortRange,
```
taspelund: Same feedback here. I'd like to see a more descriptive arg name to make it more clear how `range` fits into `fit_mappings`

dpd/src/nat.rs
```diff
@@ -78,50 +90,39 @@ fn ipv4_entry(ipv4: Ipv4Addr, e: &Ipv4NatEntry) -> String {
...
         .collect()
 }
 
-fn find_space<T: PortRange>(
-    entries: &[T],
-    low: u16,
-    high: u16,
+fn find_space(
+    ranges: impl ExactSizeIterator<Item = PortRange>,
+    range: PortRange,
```
taspelund: ditto here. I think this function stands to gain the most from an arg rename, because we iterate over `ranges` looking for `range`... and that's a bit ambiguous.  At least to me it is

taspelund (changes requested):

dpd/src/nat.rs
nicolaskagami: I see what you mean. The `ranges` are the ones currently there while `range` is the one to find space for (I'm looking forward to changing this in a future rework). Would `range_to_place` sound good to you?

taspelund: I would lean towards something closer to `candidate` or `candidate_range`. However, a small comment could also help make the name less significant

dpd/src/nat.rs
internet-diglett: I removed this check and ran the test suite and the `test_mapping` test case still passed (the test I made a different comment about):

```diff
diff --git a/dpd/src/nat.rs b/dpd/src/nat.rs
index 51e9909..ad2b5ce 100644
--- a/dpd/src/nat.rs
+++ b/dpd/src/nat.rs
@@ -115,15 +115,13 @@ fn find_space(
     range: PortRange,
 ) -> Option<usize> {
     let len = ranges.len();
-    let mut iter = ranges.enumerate().peekable();
+    let mut iter = ranges.enumerate();

     while let Some((idx, e)) = iter.next() {
         if e.overlaps(range) {
             return None;
         }
-        if e.low >= range.high
-            && iter.peek().is_none_or(|(_, next)| next.low >= range.high)
-        {
+        if e.low >= range.high {
             return Some(idx);
         }
     }
```

```
cargo nextest run test_mapping --features=tofino_stub
   Compiling dpd v0.2.0 (/disk1/workspace/dendrite/dpd)
    Finished `test` profile [unoptimized + debuginfo] target(s) in 26.87s
────────────
 Nextest run ID c112b810-a8fd-4413-be7b-36375c282d8e with nextest profile: default
    Starting 1 test across 20 binaries (81 tests skipped)
        PASS [   0.017s] dpd::bin/dpd nat::test_mapping
────────────
     Summary [   0.018s] 1 test run: 1 passed, 81 skipped
```

Which led me to believe one of the following _may_ be true:

* The check is implicitly satisfied by some other behavior in the code
* The test is not actually exercising that check (which we should fix if true)

So I looked deeper to see if there was something else I was not understanding.

From what I can tell from checking each call site, we use the index returned from `find_space` to determine where to insert a given `PortRange` into an existing collection of `PortRange`s called `ranges`. Due to this, this collection seems to be implicitly sorted. In fact, if you give `find_space` an unsorted set of `ranges` it will silently break (I verified this by adjusting the test cases, so this might also be worth adding a comment about or putting a guard in place).

Using the test entries [1-4, 7-10, 12-18] with range = [11-11]:

  - idx=0, e=[1-4]: no overlap; e.low(1) >= range.high(11)? No → continue
  - idx=1, e=[7-10]: no overlap; e.low(7) >= range.high(11)? No → continue
  - idx=2, e=[12-18]: no overlap; e.low(12) >= range.high(11)? Yes → peek is None (no next entry) → return Some(2)

  Now the peek case: using range = [5-6] against the same entries:

  - idx=0, e=[1-4]: no overlap; e.low(1) >= range.high(6)? No → continue
  - idx=1, e=[7-10]: no overlap; e.low(7) >= range.high(6)? Yes → peek is [12-18], next.low(12) >= range.high(6)? Yes → return Some(1)

In that second example, the peek fires and the subsequent comparison evaluates to true... and I think that subsequent comparison will _always_ be true because we're always passing in a sorted set of entries (and we must pass in a sorted set of entries or the function silently breaks). The peek and comparison can never return false here. If you give `find_space` and unsorted collection of `entries`, this check will not protect us there either.

dpd/src/nat.rs
internet-diglett: 🤔 Maybe `existing_ranges`, `existing_range`, and `new_range`?

Seems to read well in a comment like "finds a location to insert a `new_range` into a collection of `existing_ranges`"

dpd/src/nat.rs
nicolaskagami: Yep. It's definitely assuming its sorted (and ensuring it, currently). 

The one case where that check would be useful is the following:
- Insert [1-3]
- Insert [7-9]
- Try to insert [4-8]. It will check that the next one, if it exists, has its low below the high of the current range to insert. In that case, it wouldn't fit and it would correctly return false.

You're correct that the tests don't check this at the moment. We can fix that easily.

Overall, as I mentioned elsewhere, I'd really like to rework this whole thing in the future such that these invariants are explicit.

internet-diglett: I agree, explicit invariants, where possible, would make a lot of this much easier to reason about. These edge cases are also why I would like to cover more of the networking logic with property based testing as well.

internet-diglett: > Try to insert [4-8]. It will check that the next one, if it exists, has its low below the high of the current range to insert. In that case, it wouldn't fit and it would correctly return false.

Would this not get caught by the `overlap` check?

nicolaskagami: Yeah. That would be more readable, though technically one of the conditions in `overlap` is impossible in this case. Maybe that's what the original author had in mind? There are definitely lower-hanging fruit for performance.

internet-diglett: What I'm referring to is that we are _already_ performing the overlap check in the lines above this check:

```diff
diff --git a/dpd/src/nat.rs b/dpd/src/nat.rs
index 51e9909..ed1c7b9 100644
--- a/dpd/src/nat.rs
+++ b/dpd/src/nat.rs
@@ -115,21 +115,29 @@ fn find_space(
     range: PortRange,
 ) -> Option<usize> {
     let len = ranges.len();
-    let mut iter = ranges.enumerate().peekable();
+    let mut iter = ranges.enumerate();

     while let Some((idx, e)) = iter.next() {
         if e.overlaps(range) {
             return None;
         }
-        if e.low >= range.high
-            && iter.peek().is_none_or(|(_, next)| next.low >= range.high)
-        {
+        if e.low >= range.high {
             return Some(idx);
         }
     }
     Some(len)
 }

+#[test]
+fn test_overlap() {
+    let entries =
+        [PortRange::new(1, 3).unwrap(), PortRange::new(7, 9).unwrap()];
+
+    let space =
+        find_space(entries.iter().copied(), PortRange::new(4, 8).unwrap());
+    assert!(space.is_none())
+}
+
 #[test]
 fn test_mapping() {
     let entries = [
```

```
cargo nextest run test_overlap --features=tofino_stub
   Compiling dpd v0.2.0 (/disk1/workspace/dendrite/dpd)
    Finished `test` profile [unoptimized + debuginfo] target(s) in 27.39s
────────────
 Nextest run ID 24d97d46-0945-4de6-95f6-9f7a8163d550 with nextest profile: default
    Starting 1 test across 20 binaries (82 tests skipped)
        PASS [   0.013s] dpd::bin/dpd nat::test_overlap
────────────
     Summary [   0.014s] 1 test run: 1 passed, 82 skipped
```

nicolaskagami: You're right. `e` is already the "next" one and that line is testing the next-next one. `overlaps` tested the previous one.

It's coming off and I'm adding a line to the tests.

dpd/src/nat.rs
nicolaskagami: Dedided for `range_to_find` for the `find*mapping*` functions and `candidate_range` for `find_space`.

Addressed in [25b94af](https://github.com/oxidecomputer/dendrite/commit/25b94aff0a02e4363c1a2b6123943cee427297ac)

dpd/src/nat.rs
nicolaskagami: Addressed in [25b94af](https://github.com/oxidecomputer/dendrite/commit/25b94aff0a02e4363c1a2b6123943cee427297ac)

dpd/src/nat.rs
nicolaskagami: Addressed in [25b94af](https://github.com/oxidecomputer/dendrite/commit/25b94aff0a02e4363c1a2b6123943cee427297ac).

dpd/src/nat.rs
internet-diglett: NIce. I know we want to avoid a "Chesterton’s Fence" situation so I'm happy we were able to beat up on this with some test cases to try to reason about what it was or was not doing. Thanks for looking at this with me.

taspelund (approved):
LGTM, although I'd like to see an approval from @internet-diglett as well before merging

internet-diglett (approved):

PR #335, opened 2026-08-06
[2/3] [nat] introduce `NatAddress` trait

nicolaskagami:
This PR:
- Introduces a `NatAddress` trait, tying each IP address family to its p4 table, match key, and action types. Replaces duplicated per-family entry points.
- Fixes a bug where we'd clear our internal representation of the mappings if the underlying code failed to reset.

This is the second of 3 PRs simplifying and de-duplicating some of the `nat.rs` code.
- #334
- #335
- #336

Obs: Changes are almost entirely equivalent, except for the ordering of some things and the log message `nat tables` -> `nat table`.

### DIFF
--- a/dpd/src/nat.rs
+++ b/dpd/src/nat.rs
@@ -11,7 +11,8 @@ use std::net::{IpAddr, Ipv4Addr, Ipv6Addr};
 use std::ops::Bound;
 
 use crate::Switch;
-use crate::table::nat;
+use crate::table;
+use crate::table::nat::{add_entry, delete_entry};
 use crate::types::{DpdError, DpdResult};
 use common::nat::{Ipv4Nat, Ipv6Nat};
 use common::network::NatTarget;
@@ -44,6 +45,14 @@ impl PortRange {
 
     fn overlaps(self, other: PortRange) -> bool {
         self.low <= other.high && self.high >= other.low
+    }
+
+    pub(crate) fn low(self) -> u16 {
+        self.low
+    }
+
+    pub(crate) fn high(self) -> u16 {
+        self.high
     }
 }
 
@@ -277,7 +286,7 @@ pub fn set_ipv6_mapping(
         }
     };
 
-    match nat::add_ipv6_entry(switch, nat_ip, low, high, tgt) {
+    match add_entry(switch, nat_ip, l4_ports, tgt) {
         Err(e) => {
             error!(switch.log, "failed to add {}: {:?}", full, e);
             Err(e)
@@ -311,12 +320,7 @@ pub fn clear_ipv6_mapping(
             nat.ipv6_mappings.remove(&nat_ip);
         }
         let full = ipv6_entry(nat_ip, &ent);
-        return match nat::delete_ipv6_entry(
-            switch,
-            nat_ip,
-            ent.l4_ports.low,
-            ent.l4_ports.high,
-        ) {
+        return match delete_entry(switch, nat_ip, ent.l4_ports) {
             Err(e) => {
                 error!(switch.log, "failed to clear {}: {:?}", full, e);
                 Err(e)
@@ -452,7 +456,7 @@ pub fn set_ipv4_mapping(
         }
     };
 
-    match nat::add_ipv4_entry(switch, nat_ip, low, high, tgt) {
+    match add_entry(switch, nat_ip, l4_ports, tgt) {
         Err(e) => {
             error!(switch.log, "failed to add nat entry {}: {:?}", full, e);
             Err(e)
@@ -501,12 +505,7 @@ pub fn clear_ipv4_mapping(
             nat.ipv4_mappings.remove(&nat_ip);
         }
         let full = ipv4_entry(nat_ip, &ent);
-        return match nat::delete_ipv4_entry(
-            switch,
-            nat_ip,
-            ent.l4_ports.low,
-            ent.l4_ports.high,
-        ) {
+        return match delete_entry(switch, nat_ip, ent.l4_ports) {
             Err(e) => {
                 error!(switch.log, "failed to clear {}: {:?}", full, e);
                 Err(e)
@@ -561,12 +560,7 @@ pub fn clear_overlapping_mappings_v4(
         for idx in mappings_to_delete {
             let ent = mappings.remove(idx);
             let full = ipv4_entry(nat_ip, &ent);
-            match nat::delete_ipv4_entry(
-                switch,
-                nat_ip,
-                ent.l4_ports.low,
-                ent.l4_ports.high,
-            ) {
+            match delete_entry(switch, nat_ip, ent.l4_ports) {
                 Err(e) => {
                     error!(switch.log, "failed to clear {}: {:?}", full, e);
                     return Err(e);
@@ -606,12 +600,7 @@ pub fn clear_overlapping_mappings_v6(
         for idx in mappings_to_delete {
             let ent = mappings.remove(idx);
             let full = ipv6_entry(nat_ip, &ent);
-            match nat::delete_ipv6_entry(
-                switch,
-                nat_ip,
-                ent.l4_ports.low,
-                ent.l4_ports.high,
-            ) {
+            match delete_entry(switch, nat_ip, ent.l4_ports) {
                 Err(e) => {
                     error!(switch.log, "failed to clear {}: {:?}", full, e);
                     return Err(e);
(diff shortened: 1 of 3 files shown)


dpd/src/nat.rs
```diff
@@ -277,7 +277,7 @@ pub fn set_ipv6_mapping(
         }
     };
 
-    match nat::add_ipv6_entry(switch, nat_ip, low, high, tgt) {
+    match nat_ip.add_entry(switch, low, high, tgt) {
```
internet-diglett: 🤔

* We're validating the port range with a newtype but not using it here
* We're also passing in the `NatTarget`
* Both the `PortRange` and the `NatTarget` were used to construct the `Ipv6NatEntry` type earlier

Does it make sense to pass the `Ipv6NatEntry` here as well? It wouldn't surprise me if there were additional validations we could perform with that type as well.

Same sentiment with other similar calls.

dpd/src/table/nat.rs
```diff
@@ -18,16 +19,103 @@ use crate::Switch;
...
+
+    fn reset(s: &Switch) -> DpdResult<()> {
+        debug!(s.log, "resetting {} nat table", Self::NAME);
+        s.table_clear(Self::TABLE).inspect_err(|e| {
+            error!(s.log, "failed to reset {} nat table: {:?}", Self::NAME, e);
+        })
+    }
+}
+
+impl NatAddress for Ipv4Addr {
```
internet-diglett, line 76: Instead of implementing `NatAddress` for a raw V4 or V6 addr, would it make sense to create newtypes around them that provides additional validation (i.e. do we want to allow creation of a NAT entry for _any_ Ipv4 / Ipv6 address?), and implement this trait for the newtype?

nicolaskagami: We could, but I didn't want to add new functionality on these PRs.

dpd/src/nat.rs
nicolaskagami: - Some of what we're doing here is in service of #336. This is one place where it could hold this type but it wouldn't be of much use since we would immediately de-structure it, effectively adding another type to the trait just so we could de-structure it one level below.
- This PR was as semantic-preserving as possible and avoided bigger refactors. I actually think `PortRange` could be separated off and we could have a more bespoke struct / search algorithms for it (as mentioned on the description of #334).

nicolaskagami: I experimented with passing in `PortRange` and I think it's slightly more readable. Addressed in [1b86758](https://github.com/oxidecomputer/dendrite/commit/1b86758df72b36745dc68e9426d3517e55dc67f6) (sorry about the rebases).

dpd/src/nat.rs
```diff
@@ -632,27 +620,15 @@ pub fn clear_overlapping_mappings_v6(
...
 
-    debug!(switch.log, "resetting ipv6 nat tables");
     nat.ipv6_mappings.clear();
-    if let Err(e) = nat::reset_ipv6(switch) {
-        error!(switch.log, "failed to reset ipv6 nat table: {:?}", e);
-        Err(e)
-    } else {
-        Ok(())
-    }
+    Ipv6Addr::reset(switch)
```
internet-diglett: 🤔 This feels less clear than the older representation. If `nat::reset_ipv6` was called in the wrong context it would be clear. If `Ipv6Addr::reset(switch)` is called in any other context it would not be clear if we're doing the correct thing or not without looking into the function to see what it did.

I _feel_ like this is a close-but-slightly-awkward abstraction and this should be something more akin to an `Ipv6NatTable` type that implements a `NatTable` trait, or something in that spirit. I can be convinced otherwise, but even after sleeping on it the proposed representation feels weird to me.

dpd/src/nat.rs
```diff
@@ -632,27 +620,15 @@ pub fn clear_overlapping_mappings_v6(
...
 
-    debug!(switch.log, "resetting ipv4 nat tables");
     nat.ipv4_mappings.clear();
-    if let Err(e) = nat::reset_ipv4(switch) {
-        error!(switch.log, "failed to reset ipv4 nat table: {:?}", e);
-        Err(e)
-    } else {
-        Ok(())
-    }
+    Ipv4Addr::reset(switch)
```
internet-diglett: same as above

dpd/src/table/nat.rs
internet-diglett: additional validations could be added later

internet-diglett (changes requested):

dpd/src/nat.rs
nicolaskagami: I guess I split this and #336 in an unfortunate place since this does go away in the next PR. Should be easy to address it on this one though.

nicolaskagami: Addressed in [8293a3e](https://github.com/oxidecomputer/dendrite/commit/8293a3ed645daf42e85bd1e18fd9092e1fd46b18). Definitely better, thanks.

dpd/src/table/nat.rs
nicolaskagami: I really feel like such a structure should be added on a dedicated PR, where we can discuss the semantics we want to enforce.

dpd/src/nat.rs
```diff
@@ -632,27 +621,15 @@ pub fn clear_overlapping_mappings_v6(
...
 
-    debug!(switch.log, "resetting ipv6 nat tables");
     nat.ipv6_mappings.clear();
-    if let Err(e) = nat::reset_ipv6(switch) {
-        error!(switch.log, "failed to reset ipv6 nat table: {:?}", e);
-        Err(e)
-    } else {
-        Ok(())
-    }
+    table::nat::reset::<Ipv6Addr>(switch)
```
bnaecker: This predates your PR, but it might make sense to swap the order of operations here. If we fail to delete the entries on switch table, we'll have already cleared the local mappings, and have some torn state. Could we move the `nat.ipv6_mappings.clear()` call after a successful call to `table::nat::reset`?

dpd/src/nat.rs
```diff
@@ -632,27 +621,15 @@ pub fn clear_overlapping_mappings_v6(
...
 
-    debug!(switch.log, "resetting ipv4 nat tables");
     nat.ipv4_mappings.clear();
-    if let Err(e) = nat::reset_ipv4(switch) {
-        error!(switch.log, "failed to reset ipv4 nat table: {:?}", e);
-        Err(e)
-    } else {
-        Ok(())
-    }
+    table::nat::reset::<Ipv4Addr>(switch)
```
bnaecker: Same here, I'd argue we should swap clearing the in-memory and switch-table mappings.

dpd/src/table/nat.rs
```diff
@@ -7,27 +7,122 @@
...
+}
+
+pub(crate) fn table_dump<A: NatAddress>(
+    s: &Switch,
+    from_hardware: bool,
+) -> DpdResult<table::Table> {
+    s.table_dump::<A::MatchKey, A::Action>(A::TABLE, from_hardware)
+}
+
+pub(crate) fn counter_fetch<A: NatAddress>(
```
bnaecker: Small nit, but I think the counter- and dump-related functions can be private to the `table` mod. They're only called from `dpd/src/table/mod.rs` (e.g., line 333), not more generally in the `dpd` crate.

dpd/src/table/nat.rs
```diff
@@ -7,27 +7,122 @@
...
 use crate::Switch;
+use crate::nat::PortRange;
 use crate::table::*;
 use common::network::{MacAddr, NatTarget};
 
+pub(crate) trait NatAddress: Copy + Ord + fmt::Display {
+    const TABLE: TableType;
+    const NAME: &'static str;
+
+    type MatchKey: MatchParse + Hash + fmt::Display;
```
bnaecker, line 27: I'm probably missing something, but where do we use the `Hash` bound?

bnaecker (approved):
Looks good, nice cleanup. Just a few questions / nits.

dpd/src/table/nat.rs
bnaecker: EDIT: Oh I see, this is required in methods like `Switch::table_entry_del()` and ultimately `aal::TableOps::entry_del()`. I wonder if that's really needed, but it's not worth fixing on this PR.

dpd/src/table/nat.rs
nicolaskagami: Yeah, they could be `pub(super)`

dpd/src/nat.rs
nicolaskagami: Good catch. Definitely an underlying bug I can happily fix. This could've made the mapping representation drift from the actual switch state.

nicolaskagami: Addressed in [75f67a5](https://github.com/oxidecomputer/dendrite/commit/75f67a5a1c242fadee2c9f11f7aba698999a4e70)

dpd/src/nat.rs
nicolaskagami: Addressed in [75f67a5](https://github.com/oxidecomputer/dendrite/commit/75f67a5a1c242fadee2c9f11f7aba698999a4e70)

dpd/src/table/nat.rs
nicolaskagami: Addressed in [75f67a5](https://github.com/oxidecomputer/dendrite/commit/75f67a5a1c242fadee2c9f11f7aba698999a4e70)

dpd/src/table/nat.rs
nicolaskagami: Opened #338 for this.

internet-diglett (approved):
This looks really good, thank you for working on this!

PR #336, opened 2026-08-06
[3/3] [nat] unify per-family NAT mapping

nicolaskagami:
This PR:
- Introduces a `NatEntry` and `NatMap` to replace duplicated v4/v6 logic (net -203 LOC).

This is the last of 3 PRs simplifying and de-duplicating some of the `nat.rs` code.
- #334
- #335
- #336

Obs: 
- I decided to upgrade the IPv6 log error from `trace` to `error` to match the IPv4 one but I'm happy to do it the other way around.
- Renamed `ipv4_generation` to `generation`, since it applies to both even though the nexus API nomenclature is out-of-date.


dpd/src/nat.rs
```diff
@@ -178,227 +153,115 @@ fn test_mapping() {
...
-            }
-            _ => {
-                debug!(switch.log, "cleared nat entry {}", full);
-                Ok(())
-            }
-        };
-    }
+pub(crate) fn generation(switch: &Switch) -> i64 {
+    let data = switch.nat.lock();
+    debug!(switch.log, "fetching nat generation");
```
bnaecker: Nit: I don't know how often we ask for the generation, but this seems like a `trace!()` message IMO.

dpd/src/nat.rs
```diff
@@ -178,227 +153,115 @@ fn test_mapping() {
...
+    data.generation
+}
 
-    Ok(())
+pub(crate) fn set_generation(switch: &Switch, generation: i64) {
+    let mut data = switch.nat.lock();
+    debug!(switch.log, "setting nat generation");
```
bnaecker: Same nit here. And either way, I'd add the generation we're setting to here in the log message, e.g., `debug!(switch.log, "setting nat generation"; generation);` or similar.

dpd/src/nat.rs
```diff
@@ -178,227 +153,115 @@ fn test_mapping() {
...
-) -> Vec<Ipv4Nat> {
-    max = std::cmp::min(max, 64);
-    let nat = switch.nat.lock().unwrap();
-    let mappings = match nat.ipv4_mappings.get(&external) {
-        Some(m) => m,
-        None => return Vec::new(),
-    };
+    max: usize,
+) -> Vec<A::Reservation> {
+    let max = max.min(64);
```
bnaecker, line 233: This predates your PR, but this is pretty surprising! We'll ignore the client's requested max page size, clamping it to 64. I feel like this needs another look, probably worth an issue?

dpd/src/nat.rs
```diff
@@ -469,42 +307,41 @@ pub fn set_ipv4_mapping(
...
+        IpAddr::V4(ip) => add_mapping(switch, ip, low, high, tgt),
+        IpAddr::V6(ip) => add_mapping(switch, ip, low, high, tgt),
     }
 }
 
-/// Find the first `NatTarget` where its `Ipv4NatEntry` matches the provided
-/// `Ipv4Addr` and overlaps with the provided port range, then remove it.
-pub fn clear_ipv4_mapping(
+/// Find the first `NatEntry` that overlaps with the provided port range,
+/// then remove it.
```
bnaecker, line 324: Not for this PR, but I think this API is a probably not quite right. IIUC, we're letting a client set a NAT mapping of, say, ports `(0, 16k)`, but delete them with an API that lists ports `(10, 11)`. That seems a little sketchy to me. I think we probably want to enforce deletion by exact ranges.

dpd/src/nat.rs
```diff
@@ -178,227 +153,115 @@ fn test_mapping() {
...
-    nat_ip: Ipv6Addr,
-    low: u16,
-    high: u16,
-    tgt: NatTarget,
-) -> DpdResult<()> {
-    let l4_ports = PortRange::new(low, high)?;
-    let new_entry = Ipv6NatEntry { l4_ports, tgt };
-    let full = ipv6_entry(nat_ip, &new_entry);
-    trace!(switch.log, "adding nat entry {}", full);
+pub struct Nat(Mutex<NatData>);
```
bnaecker, line 181: I think this move is great, hiding the mutex internally. Is it possible to go further, and put a mutex around each IP family? IIUC, we're sharing the generation, so that might not work. But it'd be nice since adding removing those mappings shouldn't impede one another AFAIK.

dpd/src/nat.rs
```diff
@@ -520,46 +357,46 @@ pub fn clear_ipv4_mapping(
...
     }
 }
 
-/// Deletes any `Ipv4NatEntry` where each entry matches the provided
-/// `Ipv4Addr` and overlaps with the provided port range
-pub fn clear_overlapping_mappings_v4(
+pub(crate) fn reset<A: NatFamily>(switch: &Switch) -> DpdResult<()> {
+    let mut data = switch.nat.lock();
+    A::family(&mut data).clear();
+    table::nat::reset::<A>(switch)
```
bnaecker: Similar to the other PR, I feel like we want to swap these. If we fail to reset the table state, we probably don't want to clear our in-memory state. I might be missing something though, caching is hard :)

bnaecker (approved):
Looks good, thanks! I have a few nits and suggestions, but they're for new issues not this PR. Nice cleanup here.

dpd/src/nat.rs
nicolaskagami: A mutex for each family would definitely fit the code better but I was carefully trying not to mess with any semantics. In this case, the generation is definitely part of it but shouldn't represent the whole of it since it's not the only source of truth, specially now with the new scrimlet reconcilers using the API to set up the service nat entries. Still, I'd take a conservative approach for now.

bnaecker: Sounds good, something to keep in mind for the future. Thanks!

dpd/src/nat.rs
nicolaskagami: The alternative would have to involve partitioning the mapping, which is definitely different semantics for the API.

dpd/src/nat.rs
nicolaskagami: I think it's healthy to have a max page size in the implementor for efficiency purposes, though 64 does sound a bit small.

dpd/src/nat.rs
nicolaskagami: Addressed in [7ee059d](https://github.com/oxidecomputer/dendrite/commit/7ee059d05d0f8289d881286847f1baaab85d44f1)

dpd/src/nat.rs
nicolaskagami: Addressed in [7ee059d](https://github.com/oxidecomputer/dendrite/commit/7ee059d05d0f8289d881286847f1baaab85d44f1)

dpd/src/nat.rs
nicolaskagami: Addressed in [7ee059d](https://github.com/oxidecomputer/dendrite/commit/7ee059d05d0f8289d881286847f1baaab85d44f1)

dpd/src/nat.rs
bnaecker: I definitely agree we should have _some_ limit. What surprised me is that it's so small, and also set at this layer of the code. We usually either let Dropshot pick the limit from the request or its configured default, or change it as needed in the HTTP handler.

I don't there's any action needed for this PR. Maybe make a TODO or issue so we can come back to it if we need. For example, this might cause problems with a bulk NAT operation API we've been discussing.

dpd/src/nat.rs
bnaecker: Definitely, it would change the semantics. In reality, I'm guessing we're only using this with "nicely" aligned port boundaries, like (0, 64k) or (0, 16k). At least, that's all Nexus will construct. In that case, having this more permissive deletion path could lead to some confusing failure modes. Like I said, nothing to do for this PR, but I would argue we want to make this stricter in the future.

dpd/src/nat.rs
```diff
@@ -178,227 +153,115 @@ fn test_mapping() {
...
-        None => return Vec::new(),
-    };
-
-    let port = match last_port {
-        None => 0,
-        Some(l) => l + 1,
-    };
+/// Ties an address family to its NAT table inside `NatData`.
+pub(crate) trait NatFamily: NatAddress {
+    fn family(data: &mut NatData) -> &mut NatMappings<Self>;
```
internet-diglett: For some reason `NatFamily` and `family()` reads weird to me. How do we feel about something like this?

```suggestion
pub(crate) trait MappedAddress: NatAddress {
    fn current_mappings(data: &mut NatData) -> &mut NatMappings<Self>;
```


This gives us

```rust
A::current_mappings(&mut data).get(&nat_ip)
```

internet-diglett (commented):
Looks really good, just one comment

dpd/src/nat.rs
nicolaskagami: how about just `mappings` then?

internet-diglett: `mappings` is also fine with me.

nicolaskagami: Addressed in [55a6051](https://github.com/oxidecomputer/dendrite/commit/55a6051fdc17caf274231be81cac7637952d3a39)

internet-diglett (approved):

PR #343, opened 2026-08-18
Move API-exposed types into dpd-types-versions

taspelund:
## Summary

- move the remaining DPD API wire types into the version module where each was introduced
- make `dpd-types`, `common`, and `aal` compatibility facades over `dpd-types-versions`
- remove non-version type dependencies from `dpd-api` and convert transceiver-controller values at the DPD runtime boundary

This is a refactor only; all blessed API documents remain unchanged.

## Testing

- `cargo test -p dpd-types-versions -p dpd-api`
- `cargo clippy -p dpd-types-versions -p dpd-types -p dpd-api -p common -p aal --all-targets`
- `cargo clippy -p dpd --all-targets --features tofino_stub`
- `cargo run --package dendrite-dropshot-apis --bin dendrite-dropshot-apis -- check` (all 12 documents fresh)


dpd-types/versions/src/initial/port.rs
```diff
@@ -7,12 +7,265 @@
...
+    Deserialize,
+    Eq,
+    Hash,
+    JsonSchema,
+    Ord,
+    PartialEq,
+    PartialOrd,
+    Serialize,
+)]
+#[serde(try_from = "String", into = "String")]
```
sunshowers, line 30: I guess since we depend on this in serde, the corresponding `TryFrom` etc impls should live here and not in `impls/`.

dpd-types/versions/src/impls/network.rs
```diff
@@ -0,0 +1,315 @@
...
+
+impl fmt::Debug for MacAddr {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(
+            f,
+            "{:02x}:{:02x}:{:02x}:{:02x}:{:02x}:{:02x}",
+            self.a[0], self.a[1], self.a[2], self.a[3], self.a[4], self.a[5]
+        )
+    }
+}
```
sunshowers: Hmm, I suspect these will be required for the v1 even if it changes, so it should live there. On the other hand I don't expect `MacAddr` to ever change :)

dpd-types/versions/src/latest.rs
```diff
@@ -173,24 +218,25 @@ pub mod snapshot {
...
 pub mod switch_port {
     pub use crate::v1::switch_port::Led;
     pub use crate::v1::switch_port::LedPolicy;
+    pub use crate::v1::switch_port::LedState;
     pub use crate::v1::switch_port::ManagementMode;
     pub use crate::v1::switch_port::SwitchPortView;
 }
 
 pub mod table {
+    pub use crate::v1::counters::CounterData;
```
sunshowers: I'd drop this re-export since it's already done by `counters`.

sunshowers (commented):
Looks great, just a few comments.

taspelund:
Thanks @sunshowers, all items addressed!

sunshowers (approved):

PR #346, opened 2026-08-18
Add `allow_ddm_traffic` parameter for link creation

andrewjstone:
Plumb this through such front ports can run ddm for multirack.


dpd-types/versions/src/allow_ddm_traffic/link.rs
```diff
@@ -0,0 +1,75 @@
...
+
+use common::ports::{PortFec, PortSpeed, TxEq};
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
+
+use crate::v1;
+
+/// Parameters used to create a link on a switch port.
+#[derive(Clone, Debug, Deserialize, Serialize, JsonSchema)]
+pub struct LinkCreate {
```
taspelund, line 15: We decided against renaming `LinkCreate` for now?

dpd-types/versions/src/allow_ddm_traffic/link.rs
```diff
@@ -0,0 +1,75 @@
...
+    /// Transceiver equalization adjustment parameters.
+    /// This defaults to `None`.
+    #[serde(default)]
+    pub tx_eq: Option<TxEq>,
+
+    /// Whether DDM traffic is allowed on this link.
+    ///
+    /// This defaults to `false`.
+    #[serde(default)]
+    pub allow_ddm_traffic: bool,
```
taspelund, line 53: why carry this in `LinkCreate` instead of in `LinkSettings`?

dpd-types/versions/src/allow_ddm_traffic/link.rs
andrewjstone: Yeah, we definitely talked about this in this morning's meeting. The idea was that larger changes will end up reworking this with everything else. I'm only to make the minimum change right now.

dpd/src/link.rs
```diff
@@ -337,6 +337,9 @@ pub struct LinkConfig {
     /// This link is expected to be connected to the outside world, and
     /// should only accept inbound traffic that matches a NAT mapping.
     pub uplink: bool,
+
+    /// This links should allow ddm traffic for multirack setups
```
taspelund: typo: `s/links/link/`

dpd-types/versions/src/allow_ddm_traffic/link.rs
andrewjstone: Because we want it to work not just for `port_settings_apply`, but for `link_create`. This is the least invasive way to do that. We discussed it and said the types will probably change in the near future anyway, and so adding another wrapper seemed like overkill.

taspelund (approved):
Looks pretty straightforward to me. Just one typo and a couple questions about struct naming + which struct owns the new param.

dpd-api/src/lib.rs
```diff
@@ -1645,11 +1667,38 @@ pub trait DpdApi {
...
 
+    /**
+     * Apply port settings atomically.
+     *
+     * These settings will be applied holistically, and to the extent possible
+     * atomically to a given port. In the event of a failure a rollback is
+     * attempted. If the rollback fails there will be inconsistent state. This
+     * failure mode returns the error code "rollback failure". For more details see
+     * the docs on the [`PortSettings`] type.
+     */
```
cfzimmerman, line 1678: Is there a convention for when we use `/** ... */`  versus `///` doc comments? This module has both. Aesthetically, sticking to one is nice. Feel like I don't see `/** */` in the wild very often.

But you don't need to change this. Just curious.

cfzimmerman: That said, the contents of the comment are great. When thinking of future design, it would be nice to never need "inconsistent state" in documentation :sweat_smile:

cfzimmerman (approved):
Lgtm!

I appreciate this is time sensitive, so feel free to merge. However I'm curious about validation precedent.

Given more time, how might we test this? What do you think is warranted validation for a change like this? Here are some ideas:

- A basic dpd-client test could apply these settings and verify that `uplink == true` on the resulting link. This is pretty low value, but I guess it would help something like a messy git resolution or big refactor.
- Do we have any existing regression tests on the omicron/dendrite API? Guessing these would live on the omicron side?
- I feel that we should eventually have an automated CI test somewhere for multirack networking. Since it combines (at least) omicron, maghemite, and dendrite systems, would we orchestrate this on a4x2/voxel? Do we have precedent/infra for this kind of automated testing?

dpd-api/src/lib.rs
andrewjstone: This comment was part of the original docs and just copied verbatim to the new version of the API. I personally think we should use `///` everywhere. But in this case it didn't make sense for me to modify it.

andrewjstone:
> Lgtm!
> 
> I appreciate this is time sensitive, so feel free to merge. However I'm curious about validation precedent.
> 
> Given more time, how might we test this? What do you think is warranted validation for a change like this? Here are some ideas:
> 
>     * A basic dpd-client test could apply these settings and verify that `uplink == true` on the resulting link. This is pretty low value, but I guess it would help something like a messy git resolution or big refactor.
> 
>     * Do we have any existing regression tests on the omicron/dendrite API? Guessing these would live on the omicron side?
> 
>     * I feel that we should eventually have an automated CI test somewhere for multirack networking. Since it combines (at least) omicron, maghemite, and dendrite systems, would we orchestrate this on a4x2/voxel? Do we have precedent/infra for this kind of automated testing?

Thanks for taking a look Cory. I'm in the process of testing this in voxel, although it requires a few more changes in omicron to do so. I'm not going to merge until all the related PRs are ready to go.

Besides that I have no idea how to test this. I don't know what tests do or don't exist and what people normally do for changes like this.

cfzimmerman:
> Besides that I have no idea how to test this. I don't know what tests do or don't exist and what people normally do for changes like this.

Sgtm, thx for clarifying. Figuring this out and adding any documentation or tooling to standardize dendrite PR validation is on my to-do list.

dpd/src/main.rs
```diff
@@ -767,6 +767,7 @@ async fn sidecar_main(mut switch: Switch) -> anyhow::Result<()> {
                         kr: true,
                         lane: Some(dpd_types::link::LinkId(0)),
                         tx_eq: None,
+                        allow_ddm_traffic: false,
```
nicolaskagami, line 770: Shouldn't DDM run on rear ports?

dpd/src/api_server.rs
```diff
@@ -2963,6 +2963,7 @@ impl From<&crate::link::Link> for LinkSettings {
                 autoneg: l.config.autoneg,
                 kr: l.config.kr,
                 tx_eq: l.tx_eq,
+                allow_ddm_traffic: false,
```
nicolaskagami: Shouldn't we pass it through here?
```suggestion
                allow_ddm_traffic: l.config.allow_ddm_traffic,
```

dpd-types/versions/src/allow_ddm_traffic/link.rs
```diff
@@ -0,0 +1,75 @@
...
```
nicolaskagami, line 53: I think the semantics are not matching the behavior. `allow_ddm_traffic: false` still allows DDM traffic is it isn't a `Qsfp` port. We should either change the description (and name) to match the current behavior or the other way around.

dpd/src/port_settings.rs
```diff
@@ -102,6 +102,7 @@ struct LinkSpec {
     pub ipv4: BTreeSet<Ipv4Addr>,
     pub ipv6: BTreeSet<Ipv6Addr>,
     pub tx_eq: Option<TxEq>,
+    pub allow_ddm_traffic: bool,
```
nicolaskagami, line 105: We're adding a new field here but the `modify_link` wasn't changed to apply it. It would definitely benefit from a bit of a refactor, or at least a destructuring pattern, so we'd get compilation errors next time this happens.

dpd/src/link.rs
```diff
@@ -417,10 +421,18 @@ impl Link {
...
+        let ipv6_enabled =
+            !matches!(port_id, PortId::Qsfp(_)) || params.allow_ddm_traffic;
         // By default we expect external-facing links to be used as uplinks and
         // internal-facing links for backplane traffic.
-        let uplink = matches!(port_id, PortId::Qsfp(_));
+        //
+        // We need to disable uplinks for multirack ddm traffic on the front
+        // ports.
+        let uplink =
+            matches!(port_id, PortId::Qsfp(_)) && !params.allow_ddm_traffic;
```
nicolaskagami, line 438: Is this the only place where `allow_ddm_traffic` effects change? If so, it seems to only affect `Qsfp` ports, which is different semantics than what it's described as.

nicolaskagami (changes requested):
Found a few problems that need looking into:
- Easy: new `LinkSpec` field but no `modify_link` change.
- Easy: I think we want to pass it through in `api_server.rs`, no?
- Annoying: I thought it was wrong to set it to `false` for rear ports, but as I mentioned on another comment I think the problem is that the documented semantics don't match the implementation. It currently only seems to apply for `Qsfp`, more specifically just setting Ipv6 and disabling `uplink`. At the very least we need to communicate this better for maintainability.

nicolaskagami:
Applied the two "easy" fixes (after talking to @andrewjstone). There's a bunch around that could be improved long-term. Particularly:
- There's some duplication of business logic going on.
- The whole `Modify<_>` could be simplified and made more robust.

andrewjstone:
Thanks for the fixes @nicolaskagami. Much appreciated.

andrewjstone:
@nicolaskagami I think I updated all the relevant comments to reflect reality.

nicolaskagami (approved):
Thanks for bearing with me!

dpd-api/src/lib.rs
```diff
@@ -721,6 +723,25 @@ pub trait DpdApi {
         params: TypedBody<latest::link::LinkCreate>,
     ) -> Result<HttpResponseCreated<latest::link::LinkId>, HttpError>;
```
taspelund, line 724: I think the guidance in RFD 619 is to always use fully qualified type versions in the dropshot trait, rather than `latest`.  For the endpoints you're bumping, I would also change the signature to use the specific version in the path i.e. `v13::link::LinkCreate`

andrewjstone: We only use specif versions for prior prior versions of the API. The latest version should [use a floating identifier](https://rfd.shared.oxide.computer/rfd/0619#determinations-api-latest-floating) .

andrewjstone: FWIW, the claude skill in omicron handled this correctly.

andrewjstone:
Good news, this appears to work correctly. With a small change to voxel, which brings up an uplink on `qsfp0/0` and interconnect port on `qsfp2/0`, we see the settings correctly set in dpd.

Here's the relevant voxel patch in `voxel/src/rss_request.rs`:

```diff
@@ -132,6 +133,7 @@ fn uplink_port(p: &UplinkPort, mode: RouterMode) -> Result<PortConfig> {
         autoneg: false,
         lldp: Some(lldp(&p.switch, &p.lldp)),
         tx_eq: None,
+        allow_ddm_traffic: false,
     })
 }

@@ -152,6 +154,7 @@ fn interconnect_port(switch: &str, port: &str) -> Result<PortConfig> {
         autoneg: false,
         lldp: Some(lldp(switch, &format!("interconnect-{port}"))),
         tx_eq: None,
+        allow_ddm_traffic: true,
     })
 }
 ```

That triggers the `PortConfig` in the `RackNetworkConfig` for those ports to be send to RSS which puts them in the bootstore. The dpd-reconciler reads them and calls `port_settiings_apply` on the dpd client. 

Looking at the output after RSS we see:

```
root@oxz_switch:~# swadm link get-prop qsfp0/0 uplink
true
root@oxz_switch:~# swadm link get-prop qsfp0/0 ipv6-enabled
false
root@oxz_switch:~# swadm link get-prop qsfp2/0 uplink
false
root@oxz_switch:~# swadm link get-prop qsfp2/0 ipv6-enabled
true
```

Issue #145, opened 2025-10-03
`swadm link serdes set tx-eq` should require at least one gain value


bnaecker:
See:

https://github.com/oxidecomputer/dendrite/blob/b3cdd095437f388069aa57be5efac4b4fadc7b4f/swadm/src/link.rs#L214-L233

The comment says that `swadm link serdes set tx-eq` requires at least one value for the gains. That's a good idea, but the code actually lets you set none of them, in which case they're all defaulted to zero. We should disallow that, and require at least one.

rcgoodfellow:
What about when the user wants to explicitly set everything to `None` to pick up the defaults?

PR #350, opened 2026-08-22
Regression tests for swadm

cfzimmerman:
> "_A network switch is only as good as its CLI_"
>
> \- Probably someone important at some point

`swadm` is a great tool, and I would like it to continuously improve. While working on https://github.com/oxidecomputer/dendrite/issues/145, I observed how easy (and tempting!) it is to make breaking changes.  However, a [quick search](https://github.com/search?q=org%3Aoxidecomputer+swadm&type=code) suggests breaking changes would not be well received by scripts and docs. Such changes may be needed someday, but that should be a careful decision and certainly not an accident. 

This PR adds infrastructure for regression testing swadm changes in CI. These diffs focus on the tx eq settings I'll soon be modifying, but hopefully the structure is easy to extend as other swadm projects arise.


asic/src/tofino_common/ports.rs
```diff
@@ -212,6 +212,7 @@ impl PortData {
 /// The available_channels set is used to track the channels that have not
 /// yet been allocated.  Allocating a set of channels means that the Tofino has
 /// been configured to use those channels as a single multi-channel port.
+#[derive(Debug)]
```
cfzimmerman: todo: stray diff

cfzimmerman:
# Claude review

_(Copy/paste from local)_

# Review: dendrite#350 — swadm CLI integration tests

Nice cleanup — folding the loose `swadm/tests/*.rs` binaries into one `cli`
target, killing the dead `port-link.rs`, and giving the CLI a real
output-matching helper is the right shape. `Pattern`/`expect_line`/`pat!` is a
good abstraction and the `output_txeq` unit test that exercises the matcher
without a dpd is a nice touch.

Most of what follows is about **when** these tests run and **how** they
synchronize with dpd, not about the matcher itself. Two of them I think will
bite CI.

---

### 1. `.github/buildomat/packet-test-common.sh:106` — swadm CLI tests now gate the packet tests

Dropping `--test counters` means the swadm block runs three link-*mutating*
tests instead of one read-only one, and it runs them **before** the packet
tests, with `set -o errexit` still in effect (it's re-enabled right after the
dpd startup poll and not cleared again until `banner "Packet Tests"`).

Two consequences:

- `--no-fail-fast` only affects cargo's within-run behavior, not its exit code.
  Any swadm CLI regression now aborts the whole job before `dpd-client`'s
  integration tests run at all, so you lose all packet/multicast signal for an
  unrelated string-formatting change. That's a big blast radius for a test suite
  whose stated purpose is catching CLI drift.
- The tests mutate `rear0/0`, which `dpd-client` also depends on.
  `dpd-client/tests/integration_tests/common.rs:445` registers `rear0` in the
  port table, and `Switch::init()` (same file, line 498) **panics** —
  `failed to get mac for port rear0/0` — if `link_mac_get` fails for any
  registered port. So a swadm test that leaves `rear0/0` deleted takes down
  every packet test with a completely unrelated error message.

Suggest moving the `pushd swadm` block to *after* the `dpd-client` block. It
costs nothing and makes the swadm tests strictly additive to the job.

### 2. `swadm/tests/cli/tx_eq.rs:31` (also `link_apply.rs:24` and `:51`) — read-back races the reconciler

`link serdes set txeq` and `link apply` do **not** program the ASIC
synchronously. `Switch::link_tx_eq_set` (`dpd/src/link.rs:1277`) and the
port-settings modify path (`dpd/src/port_settings.rs:457`) both just set
`link.tx_eq`, clear `plumbed.tx_eq_pushed`, and `trigger()` the reconciler. The
read side goes the other way: `link_tx_eq_get` (`dpd/src/api_server.rs:2604`)
reads the SDE via `serdes::port_tx_eq_get`, not dpd's config.

And the reconciler doesn't just push the taps — `dpd/src/link.rs:1856`:

```rust
} else if link.config.enabled && !link.plumbed.tx_eq_pushed {
    debug!(log, "tx-eq needs an update, tearing down link",);
    true
```

so a tx-eq change **unplumbs and re-plumbs the link** (port del, port add, MAC
reprogram, tx-eq push, enable) before the new value is visible.

The tests issue the `set`/`apply` and then immediately spawn a second `swadm`
process to `get`. In the window, `serdes::port_tx_eq_get` → `lane_count()` runs
against a port that has been torn down (error → non-zero exit → `cmd::swadm`
bails), or it returns the *previous* taps (assertion failure). It'll usually
win the race because process spawn is slower than the reconciler, but "usually"
is how you get a test that fails once a month in buildomat and nobody can
reproduce.

Worth wrapping the read-back in a poll-with-timeout helper on `Out` /
`cmd::swadm` — something like `expect_line_within(Duration)` that retries the
command. Every other integration suite in this repo that touches link state
polls rather than reading once.

### 3. `swadm/tests/cli/tx_eq.rs:78` — `create_link` deletes `rear0/0` with no way back

```rust
let delete_cmd = format!("link del {link}");
if let Err(e) = cmd::swadm(&delete_cmd) { println!("Delete failed. ..."); }

cmd::swadm(format!("link create {port} -s 100g --fec rs"))?
```

If the `del` succeeds and the `create` on the next line fails — or the test
process is killed between them — `rear0/0` is gone for the rest of the job.
`dpd/misc/model_config.toml` is only read at dpd startup, so nothing puts it
back. See finding 1 for what that does to the packet tests.

A `Drop` guard (or a `scopeguard`-style teardown) that re-creates the link from
`model_config.toml`'s values (`fec = RS`, `speed = 100G`) would make these tests
safe to run anywhere in the job.

Minor related note: `create_link` recreates the link but never restores the
tx-eq state, so `set_all_taps` leaves `rear0/0` with `{-1, 0, 10, 5, 2}`
programmed. Harmless for the packet tests today since `rear0` has no veth, but
it's the kind of leftover that makes a future failure confusing.

### 4. `swadm/tests/cli/link_apply.rs:79` — `tx_eq_exclusive` only asserts "exit code != 0"

`cmd::swadm` bails on any non-zero exit, so `expect_err` can't tell "clap
rejected the conflict" from "`--lane` got renamed", "the speed string is no
longer accepted", or "dpd isn't reachable". The error already carries stderr —
asserting it contains clap's `cannot be used with` would make the test actually
pin the `conflicts_with = "tx_eq"` attributes at `swadm/src/link.rs:460-476`.

Second point on this one: this test needs no dpd at all — clap rejects the args
before any network I/O. Because it's `#[ignore]`d it only ever runs in the
Linux `packet-test` job. Un-ignoring it would get it running in the illumos
`test.sh` job for free, which is where a `swadm`-only change is most likely to
be caught early. (`#[serial]` is also unnecessary on it.)

### 5. `swadm/src/link.rs:2151` — `link apply` ignores `--link`'s link id and `--lane` (pre-existing)

Not introduced here, but the new tests are the reason to mention it: the handler
does

```rust
body.links.insert(String::from("0"), types::LinkSettings { ... });
```

hardcoding link id `0` and discarding `link.link_id` from the parsed `LinkPath`.
`--lane` is likewise dropped on the dpd side — `LinkSpec` has no `lane` field
and `add_link` (`dpd/src/port_settings.rs:325`) carries a TODO saying so.

Concretely, `swadm link apply --link rear0/1 --lane 1 --speed 100g --tag t`
reconfigures `rear0/0` and, because `calculate_links`
(`dpd/src/port_settings.rs:192`) computes `links_to_del = switch_links -
settings_links`, **deletes** `rear0/1`. Since every new test uses `rear0/0` and
`--lane 0`, the suite bakes in the blind spot. A case using a non-zero link id
would be a good addition — either as a test of the fix, or as an `#[ignore]`d
known-failure with a comment.

### 6. `swadm/tests/cli/cmd.rs:157` — a trailing `re::ANY` can never match

`make_regex` joins every pattern with `\s+`, but `expect_line` matches against
`line.trim()`. So `pat![.., ANY]` at the end of a pattern list requires trailing
whitespace that `trim()` has already removed:

```rust
out.expect_line(pat!["Speed", "100G", ANY])  // never matches `Speed  100G`
```

It happens to work at `tx_eq.rs:74` only because `link ls` has more columns
after `Media`. Since this helper is explicitly meant to be reused by future
tests, either document that `ANY` must not be last, or have `make_regex` emit
`\s*` when the next/previous element can match empty.

### 7. `swadm/tests/cli/cmd.rs:41` — hardcoded `--host [::1]` makes the CI env vars dead

`packet-test-common.sh` still exports `DENDRITE_TEST_HOST='[::1]'` and
`DENDRITE_TEST_VERBOSITY=3`, and nothing reads them any more on the swadm side.
Reading `DENDRITE_TEST_HOST` (default `[::1]`) and `DENDRITE_TEST_PORT` (default
`12224`) would keep the script honest and let these tests run against the
split illumos-dpd / Linux-model rig the new README describes — which is exactly
the setup someone will reach for when the tofino-sde#21 blocker is lifted.

### 8. `swadm/tests/cli/cmd.rs:42` — whitespace splitting is a one-way door

`split_ascii_whitespace()` makes the multi-line `format!` strings read well, but
it silently mangles any argument containing a space (a `--tag` with a space, a
description, a comma-free list). Worth an escape hatch — `swadm_args(&[&str])`
with the current function as a thin wrapper — before someone hits it and spends
an afternoon on it.

---

## Smaller things

- `swadm/tests/cli/counters.rs:17`: `.expect("Failed to execute swadm counters
  list")` now also fires when the command *ran fine and returned non-zero*.
  Wording nit, but the message will mislead whoever reads the CI log.
- `swadm/tests/cli/link_apply.rs:22`: `--tx-eq=-1` sets `main = -1`, which is
  not a physically sensible main tap (it's normally a large positive value).
  Fine against the model, which accepts anything, but if these tests ever run on
  hardware the SDE may clamp or reject it. A comment saying the value is
  deliberately arbitrary would save a future reader the detour.
- `swadm/Cargo.toml:29`: stray trailing blank line after `[dev-dependencies]`.
- README wording: "These tests are run in Linux CI but currently ignored in
  Illumos CI" is accurate, but worth adding *why* an individual test is
  `#[ignore]`d — the attribute means "needs a live dpd", and finding 4 shows at
  least one test that doesn't.

## Verified

- `cargo check -p swadm --tests` is clean on illumos; cargo picks up
  `swadm/tests/cli/main.rs` as a single `cli` test target as intended.
- `tools/check_copyrights.sh` only inspects `.sh|.xml|.rs|.p4`, so the new
  `README.md` needs no header, and all four new `.rs` files have one.
- The `tag` passed to `port_settings_apply` only lands on IP-address entries, so
  applying with `--tag test` over a link created by `link create` (or by
  `model_config.toml`) is not an ownership conflict — good.
- `dpd`'s `create_link` handles the `delete_me` case by replacing the link and
  keeping its MAC (`dpd/src/link.rs:583-605`), so the `link del` → `link create`
  sequence in `create_link` is not itself racy. It's only the failure path
  (finding 3) that's a problem.

cfzimmerman:
1. Accepted, but I don't feel great about this. If swadm tests really grow, I'd consider making them their own CI job. But that's probably its own PR.
2. Added a `cmd::repeat` helper for retries.
3. I think this is unwarranted complexity. swadm tests should assume the switch is in an undefined config and handle that accordingly (see `tx_eq::create_100g_link`). And per (1), other tests are now unaffected because swadm tests run last.

swadm/tests/cli/cmd.rs
```diff
@@ -0,0 +1,325 @@
...
+use std::borrow::Cow;
+use std::process::Command;
+use std::time::Duration;
+use std::time::Instant;
+
+use anyhow::Context;
+use anyhow::bail;
+use regex::Regex;
+
+const SWADM: &str = env!("CARGO_BIN_EXE_swadm");
```
taspelund, line 27: It looks like this is pre-existing, but do you have any idea why this is mixed case?

cfzimmerman: This was new to me as well. Seems imposed by Cargo: https://doc.rust-lang.org/cargo/reference/environment-variables.html#:~:text=cleaned%20between%20builds.-,CARGO_BIN_EXE_,-%3Cname%3E%20%E2%80%94%20The

swadm/tests/cli/cmd.rs
```diff
@@ -0,0 +1,325 @@
...
+/// The closure is not interrupted if execution exceeds timeout.
+///
+/// This is useful for the read part of write-then-read tests,
+/// where a reconciler may need some time to converge.
+pub fn retry<T, E>(mut f: impl FnMut() -> Result<T, E>) -> Result<T, E> {
+    // Modify these or make them configurable if a tested command
+    // ever requires longer than `TIMEOUT` to converge. This just
+    // avoids requiring more args if nobody cares.
+    const TIMEOUT: Duration = Duration::from_secs(2);
+    const SLEEP: Duration = Duration::from_millis(100);
```
taspelund: You could always move sleep/timeout into args and expose a thin wrapper (fn or macro) that supplies the default. Not something I feel strongly about though, so feel free to ignore me if you disagree

swadm/tests/cli/cmd.rs
```diff
@@ -0,0 +1,325 @@
...
+
+    status
+}
+
+/// Common regex patterns for searching swadm output.
+pub mod re {
+    use super::Pattern;
+
+    /// Anything up until the next match.
+    pub const ANY: Pattern = Pattern::regex(r".*?");
```
taspelund: This matches 0 or more characters, right? Is that worth calling out in the comment, since the others call out 1 being the minimum match length?

swadm/tests/cli/cmd.rs
```diff
@@ -0,0 +1,325 @@
...
+        }
+
+        Regex::new(&line_match)
+            .with_context(|| format!("Regex failed to compile: {line_match:?}"))
+    }
+}
+
+impl AsRef<str> for Output {
+    fn as_ref(&self) -> &str {
+        let (Self::Stdout(txt) | Self::Stderr(txt)) = self;
```
taspelund, line 274: this is slick, I didn't know you could do this inside of let destructuring

swadm/tests/cli/cmd.rs
```diff
@@ -0,0 +1,325 @@
...
+        // if that's warranted complexity in a test for a test for a CLI tool.
+
+        const TXEQ_STDOUT: &str = "
+           lane 0     lane 1     lane 2     lane 3
+pre2      0 (111)    1 ( 11)    2 (  1)    3 (  11)
+pre1     -1 ( 11)   -2 ( 11)   -3 ( 11)   -4 (  11)
+main     19 ( 11)   20 ( 11)   21 ( 11)   22 (  11)
+post1    -2 (  1)  -13 ( -2)   -9 (-11)  -22 (-123)
+post2  -123 ( 11)  456 ( 11)    0 ( 11)    0 (  11)
+";
```
taspelund: This feels like it would be a good use case for [expectorate](https://github.com/oxidecomputer/expectorate)

swadm/tests/cli/link_apply.rs
```diff
@@ -0,0 +1,123 @@
...
+    cmd::retry(|| {
+        let tx_eq = cmd::swadm(format!("link serdes get txeq {LINK}"))?;
+        for label in ["pre2", "pre1", "main", "post1", "post2"] {
+            tx_eq.expect_line(pat![
+                label, VAL, PARENS, VAL, PARENS, VAL, PARENS, VAL, PARENS
+            ])?;
+        }
+
+        Ok(())
+    })
```
taspelund, line 41: Similar feedback here about expectorate. As someone unfamiliar with the output of `swadm link serdes get txeq`, I have no idea what this test is looking for. I think being able to look at some output that's checked into get as a comparison would make it much simpler to reason about the test.

swadm/tests/cli/link_apply.rs
```diff
@@ -0,0 +1,123 @@
...
+                label, val, PARENS, val, PARENS, val, PARENS, val, PARENS
+            ])?;
+        }
+
+        Ok(())
+    })
+}
+
+/// Verifies that the `tx-eq` shorthand and explicit
+/// tap flags are mutually exclusive.
```
taspelund, line 72: what are tap flags? I'm not sure what this test is doing without digging into swadm to figure out what the tap flags are. Maybe a reference to a command, source file, or type name would be helpful?

taspelund (changes requested):
Overall, the changes LGTM. I would say that the main bit of feedback I have is that it's hard to track what the tests are doing without already knowing that area of the code (what is a tap flag? what does the expected output look like?).

I'd like you to take a look at [expectorate](https://github.com/oxidecomputer/expectorate) and see if that would be a good fit for some of these tests. If it doesn't make sense in this case, then maybe just a multi-line comment showing what swadm output the test is comparing against would be helpful.

swadm/tests/cli/link_apply.rs
bnaecker: This is referring to the flags like `--main` or `--pre1`.

The term "tap" comes from signal processing, and refers to one shifted copy of the signal we're operating on. So `--pre1` is the gain on a copy of the signal shifted backward in time by one sample. It's one of the gains in the filter we apply to the signal to try to improve quality.

cfzimmerman: I'm generally against unwarranted jargon. But `tap` seems pretty well standardized in tx eq literature, and it's a super convenient name :sweat_smile:

bnaecker: Yeah, the term _tap_ holds roughly the same position in signal processing as _process_ does in software. If you're mucking with it, you know what it means :)

swadm/tests/cli/cmd.rs
cfzimmerman: Haha I often write comments trying to exonerate suspicious code only to realize later the comment could just be replaced by less suspicious code. Seems the case here. Thx!

`retry` now just calls a function `retry_with` with these defaults.

swadm/tests/cli/cmd.rs
cfzimmerman: I agree that this particular case is a good use of expectorate. However since it's a test for the actual matcher, using that instead of expectorate is the whole point.

That begs whether expectorate is a better tool than `expect_line`. I'm putting a comment in the next commit, but my guess is not.

Tx eq is a good motivating case because it's arbitrarily structured and exposes uncontrolled state.

In this case, the values outside parentheses are what we have commanded, and afaik the values inside parentheses are what the hardware itself has decided to do. This occurs in the case of self optimizing tx eq. So we only want to assert the values outside parentheses controlled by swadm command. Hence the `PARENS` matches asserting that something exists there, but idk what's inside.

In my understanding, `expectorate` does exact matching, which makes it a poor fit for a fuzzy assert like this. Is that accurate?

cfzimmerman: Here's another plausible design:
- Get swadm output into a machine readable format. Either by adding formatted output options or building parsing infra.
- Allow tests to restructure that output to extract unnecessary tokens.
- Store known good versions of that output in files, and use expectorate for diffing.

However, unless someone feels strongly for a solution of that form, I'm disinclined because it seems like more required code per each new test.

swadm/tests/cli/link_apply.rs
cfzimmerman: I'm not yet convinced about expectorate, but this is a good point. I'll add an `EXAMPLE` comment.

taspelund: Sure, no worries. I don't know expectorate super well, but I've seen it used for command output checking, both structured and unstructured, so I thought it would be worth evaluating. If you don't feel like it's a good fit, that's fine by me

cfzimmerman: Sorry for the repeated pings on this. Looks like we use expectorate in a lot of other crates. If I want people to contribute tests, probably good to use familiar tooling. I'll look for a way to make it work conveniently. Thx for suggesting.

cfzimmerman (commented):
# 🤖  CLAUDE REVIEW

Nice cleanup — replacing the ad-hoc `port-link.rs` / `counters.rs` tests with a shared
`cmd` harness plus expectorate golden files is a real improvement, and the README note
about swadm being a string-typed interface is well placed.

I ran the suite against a live `dpd` + `tofino-model` rig (illumos dpd, remote model) to
check the goldens rather than just reading them. All five `--ignored` tests pass, in both
orderings I tried, and `tx_eq_exclusive` + `cmd::test::pattern_statics` pass in the
illumos `test` job's non-ignored run. So the substance is sound. The findings below are
mostly about the harness making failures hard to read and one way it can silently
weaken itself.

## The one I'd fix before merging

### `EXPECTORATE=overwrite` inside `retry` bakes in transient output

`cmd.rs:310` `try_expectorate` calls `expectorate::eq_file(..).eval(..)`, which under
`EXPECTORATE=overwrite` **writes the file and returns `true` on the first call**
(`feature_predicates.rs` → `assert_contents_impl(.., OverwriteMode::from_env(), ..)`).
Every golden comparison in this PR is wrapped in `cmd::retry`, whose whole purpose is
that the first attempt is expected to observe pre-convergence state.

Demonstrated on the rig:

```
$ EXPECTORATE=overwrite DENDRITE_TEST_HOST='[::1]' \
      cargo test --test cli set_partial_taps -- --ignored
test tx_eq::set_partial_taps ... ok

$ cat swadm/tests/cli/expect/tx_eq_100g_link.txt
Port/Link  rear0/0
Speed      100G
```

The `State      Up` line is gone. The test still passes, forever, having quietly stopped
checking that the link came up. Since regenerating goldens is exactly what someone will
do after an intentional swadm output change, this is a live trap.

Suggestion: have `retry` (or `try_expectorate`) refuse to overwrite while retrying — e.g.
retry on the *raw* command output until a cheap predicate holds, then run the expectorate
compare exactly once; or check `OverwriteMode::from_env()` and collapse to a single
attempt.

## Failure output

### A passing run prints a scary diff; a failing one prints ten

`cmd.rs:315-316` prints the expectorate diff and then `bail!("Match failed")`. Wrapped in
`retry`, every normal run of the `tx_eq` tests emits:

```
@@ -1,3 +1,2 @@
 Port/Link  rear0/0
-State      Up
 Speed      100G

string doesn't match the contents of file: ".../tx_eq_100g_link.txt" see diffset above
                set EXPECTORATE=overwrite if these changes are intentional
test tx_eq::set_partial_taps ... ok
```

That's from a *green* run — the link simply wasn't up on the first poll (this happened on
every run I did). Conversely, a genuine mismatch prints ~10 stacked diffs and then fails
with the context-free `Match failed`, which doesn't even name the file. Worth putting the
file name in the `bail!` and only printing on the last attempt.

## CI

### `packet-test-common.sh:105` — swadm tests no longer run if any packet test fails

`errexit` is on when the `dpd-client` suite runs (line 105), and the swadm block was moved
to the end of the script (lines 122-134). Before this PR the swadm checks ran first, so
both suites always reported. Now one flaky packet test aborts the script and every swadm
result disappears — the opposite of the "run all the swadm tests in CI" goal.

Moving swadm after the packet tests is right (these tests destroy `rear0/0`), so the fix
is to stop letting the first suite short-circuit the second:

```sh
set +o errexit
DENDRITE_TEST_HOST='[::1]' ... cargo test ... -- --ignored --skip succeeds_when_table_fragmented
packet_status=$?
set -o errexit
...swadm block...
exit $packet_status
```

### `tx_eq_exclusive` never runs in the Linux job

It's the only CLI test without `#[ignore]`, and the packet-test job passes `-- --ignored`,
which runs *only* ignored tests. So the one test covering clap's conflict output is
excluded from the job the new README points at; it actually runs in the illumos `test`
job (verified passing). Either mark it `#[ignore]` for consistency or add a second,
non-`--ignored` invocation.

## Smaller things

- `cmd.rs:106` — `TIMEOUT = 2s` with a 100 ms sleep gives only ~5-10 real attempts once
  you account for a process spawn plus HTTP round trip per attempt. On my idle rig
  `create_100g_link` needed a retry *every* run to see `State Up`; on the CI box, with
  `tofino-model` and the packet tests' fallout still in flight, 2 s for delete → create →
  enable → link-up is thin. Consider `retry_with` and a larger budget for the link-up wait
  specifically.

- `tx_eq.rs:71` — `create_100g_link(port, link)` hardcodes `rear0/0` in the pattern (and
  in `expect/tx_eq_100g_link.txt`) instead of using its `link` argument, so calling the
  helper with any other link fails the compare rather than checking the link asked for.

- `link_apply.rs:92` — "Ensure a consistent slate for link apply" isn't quite true for
  tx-eq. `dpd`'s `delete_link` (`dpd/src/link.rs:704`) only sets `delete_me` and pokes the
  reconciler, and the software tx-eq values survive a delete/create: after
  `link delete rear0/0`, waiting for the 404, then `link create rear0 -s 100g --fec rs`
  with no tx-eq flags, `link serdes get txeq` still reported `main -22 / post1 5` from the
  previous test. The tests pass only because both write paths (`serdes set txeq` and
  `link apply`) send all five taps. Fine today, but the comment oversells it, and any
  future test asserting a *default* tap value after create will be order-dependent.

- `tx_eq.rs:59-66` vs `link_apply.rs:93-107` — two different link-teardown helpers, one
  of which waits for the delete to land and one of which doesn't. The non-waiting one only
  works because `create_link` has an explicit "exists but marked `delete_me` → reset
  config" branch (`dpd/src/link.rs:583`). Worth collapsing into one helper in `cmd`.

- `cmd.rs:69` / `cmd.rs:79` — the doc tells the reader to use `swadm_exact` for args with
  spaces, but it's private, so sibling modules (`tx_eq`, `link_apply`) can't call it, and
  the intra-doc link points at a private item. Make it `pub`.

- `cmd.rs:87` — `"12224"` duplicates `common::DEFAULT_DPD_PORT`; use the constant.

- `swadm/Cargo.toml:30` — `predicates = "3.1.4"` is pinned in the crate while the other
  three dev-deps added here went through `[workspace.dependencies]`. Worth being
  consistent.

- Dropping `syn` / `proc-macro2` from `[workspace.dependencies]` is safe — `aal_macros`
  pins its own versions and nothing used `.workspace = true` for either. Just noting I
  checked.

cfzimmerman:
My responses:

1. I noticed as well that the link usually isn't up by the first retry. Nice motivation for the retry wrapper. I think it's okay to just look for this when reviewing swadm PRs. Version control makes the diffs visible anyway. Complicating the retry loop to avoid this seems unnecessary.
2. I again find this okay. If someone sees errors on a passing test and looks at the code, they'll find a retry loop. Seems self explanatory. Ideally we'd not emit expectorate errors during retries, but they go straight to stdout anyway.
3. Also okay with this. If the CI job is going to fail, might as well let it exit early. A PR should only merge on all greens anyway.
4. Good observation. I'll add non `--ignored` tests to this as well.
5. cmd timeout of two seconds seems alright for now. I'll look into it again if issues arise.
6. Now addressed. tx_eq and link_apply tests now use the same deletion helper, and it doesn't rely on the file constant. Made `swadm_exact` public, switched to `DEFAULT_DPD_PORT`, and moved around the dependency declaration.
7. The comments about link creation ordering rely too much on knowing the internals of DPD. And I have not verified these claims. The tests as-written treat DPD as a black box and seem sensible. I don't think they'll flake in CI. If tests fail upon iteration due to some unclear DPD design decision, that's a valid failure imo.

swadm/tests/cli/link_apply.rs
cfzimmerman: Thanks again for the feedback. Switched over to `expectorate`. Seems like a nice upgrade. And I rebased away the previous version since it's more distracting than useful now.

swadm/tests/cli/cmd.rs
```diff
@@ -0,0 +1,319 @@
...
+    /// Creates a text literal that will not engage regex semantics.
+    pub fn literal(lit: impl AsRef<str>) -> Self {
+        Regex::new(&regex::escape(lit.as_ref()))
+            .map(Self)
+            .expect("An escaped regex literal should always compile")
+    }
+
+    /// Merges an iterator of patterns into an anonymous regex
+    /// group that matches any of the patterns.
+    pub fn or(
```
cfzimmerman, line 163: The input validation could be more robust. For example, a user could make a weird regex like `(?:||)` by passing empty inputs. But I'd rather keep this simple, and`Regex::new` catches anything that's truly invalid.

swadm/tests/cli/cmd.rs
```diff
@@ -0,0 +1,319 @@
...
+    fn from(value: i32) -> Self {
+        value.to_string().into()
+    }
+}
+
+impl From<usize> for Pattern {
+    fn from(value: usize) -> Self {
+        value.to_string().into()
+    }
+}
```
taspelund: Maybe combining these into a generic impl would provide more value than just 3 concrete impls?
e.g.
```
impl<T: ToString> From<T> for Pattern {
    ...
}
```

swadm/tests/cli/cmd.rs
```diff
@@ -0,0 +1,319 @@
...
+    fn from(literal: &str) -> Self {
+        Pattern::literal(literal)
+    }
+}
+
+impl From<String> for Pattern {
+    fn from(literal: String) -> Self {
+        Pattern::literal(&literal)
+    }
+}
```
taspelund: We could maybe combine these into one generic impl and get more mileage for it too:
```
impl<T: AsRef<str>> From<T> for Pattern {
    fn from(literal: T) -> Self {
        Pattern::literal(literal.as_ref())
    }
}
```

swadm/tests/cli/cmd.rs
```diff
@@ -0,0 +1,319 @@
...
+/// and can be used for parsing CLI results.
+#[derive(Debug)]
+pub enum Output {
+    Stdout(String),
+    Stderr(String),
+}
+
+impl Output {
+    /// Removes all instances of the pattern from the output text.
+    pub fn remove(&mut self, reg: impl AsRef<Regex>) -> &mut Self {
```
taspelund: nit: perhaps a better name for this would be `strip`? when I see `remove` I think about an entry being removed from a collection or database. I also don't feel super strongly about this, so feel free to ignore this if you think the name is fine as-is

taspelund (approved):
LGTM overall. A couple nits here and there about naming or using more generics, but I'm happy with the shape.

swadm/tests/cli/cmd.rs
cfzimmerman: I agree with this in principle. In practice it's awkward because `Pattern` itself implements `AsRef<str>`, which conflicts with blanket `impl From<T> for T`.

`ToString` is also tough because `Regex` implements `ToString`, and I don't want `Pattern::from(regex)` to become a literal.

There's an elegant generic solution using marker traits, but alas that's blocked on negative impls  https://github.com/rust-lang/rust/issues/68318

I'll compromise and use a macro for now.

swadm/README.md
```diff
@@ -0,0 +1,17 @@
+# SW(itch) ADM(in)
+
+This is the management CLI for Oxide's rack switch.
+
+## Testing
+
+swadm is widely used across scripts and documentation. Changes
+should generally be backward compatible.
```
rcgoodfellow: I would not go quite this far. `swadm` is an internal development tool for `dpd`. `dpd`'s stable documented and versioned API is it's dropshot/openapi interface. While it's beneficial to have expectorate tests that make sure that PRs don't unintentionally break the CLI, backwards compatibility is not a constraint. And really, backwards compatibility of the CLI is at odds with swadm being a CLI shim to a versioned API that is meant to evolve.

cfzimmerman: Makes sense. I noticed it in [internal docs](https://github.com/search?q=org%3Aoxidecomputer++swadm&type=code) and was somewhat responding to these two PRs:
- https://github.com/oxidecomputer/dendrite/pull/352
- https://github.com/oxidecomputer/facade/pull/764

I'll change it to this:

> swadm is used across scripts and documentation. Breaking changes should be documented and intentional.

Maybe a good compromise is to start using semver.

PR #352, opened 2026-08-26
swadm: add long --speed to swadm link create

adamlouis:
Add long `--speed` option to `swadm link create` in addition to short `-s`.

The long `--speed`option was dropped in https://github.com/oxidecomputer/dendrite/pull/134

Dendrite changes were only recently integrated into `racktest` images via https://github.com/oxidecomputer/lengths-of-wire/pull/72 and forced changes in https://github.com/oxidecomputer/facade/pull/737.

Note, (1) we now have infrastructure in place to update `dendrite` in `racktest` images more easily so we should stay up to date and (2) intend to move more manufacturing sw to use Rust client libs.

### DIFF
--- a/swadm/src/link.rs
+++ b/swadm/src/link.rs
@@ -489,7 +489,7 @@ pub struct LinkCreate {
     lane: Option<types::LinkId>,
 
     /// The speed for the new link.
-    #[clap(short = 's')]
+    #[clap(short = 's', long)]
     speed: PortSpeed,
 
     /// The error-correction scheme for the link.


cfzimmerman (approved):
What you added looks reasonable. Looking at the code raises a few ideas, but none of these are required to merge. This is more a discussion about how to improve `swadm` as a tool. Would love to hear any ideas from your end as well.

- The `short = 's'` flag is not needed. [Clap will default](https://docs.rs/clap/latest/clap/_derive/index.html#:~:text=Without%20%3Cchar%3E%3A%20defaults%20to%20first%20character%20in%20the%20case%2Dconverted%20field%20name) to `s` anyway with `#[clap(short, long)]`. But this is a pattern throughout swadm, so changing it once here probably isn't worth another commit.

https://github.com/oxidecomputer/dendrite/blob/99f690f8971bcc430d26062dbdc5f0560f627358/swadm/src/link.rs#L491-L493

- In the future, I would love to see a PR like this accompanied by a corresponding test. In this case, it would both demonstrate the functionality you expect and prevent future regressions. But that's contingent upon https://github.com/oxidecomputer/dendrite/pull/350, and I'd rather not make that a merge blocker. 

- There are currently no value hints for port speed, which is poor UX. Same for `Fec` and probably other enums. I would love to slap a `clap::ValueEnum` on `PortSpeed`, but idk how polite society would receive clap macros in  `dpd-types`. (pic is not from your branch, hence no `--speed` flag)

Screenshot_2026-08-26_11-24-42

PR #356, opened 2026-08-30
[vlan] bound VLAN IDs to 1..=4094 per IEEE 802.1Q

zeeshanlakhani:
Now, we validate a `VLAN ID` against the configurable range `1..=4094`.

Per [IEEE 802.1Q §9.6](https://ieeexplore.ieee.org/document/10004498), VID 0 is the null VID used for priority tagging and carries no VLAN membership, and VID 4095 is reserved for implementation use and must not be configured. VID 1 is the default port VID.

Also route uplinkd's uplink-address and interface-name parsers through the same validation fn to match the API path.

**Note**: I'm working on a bigger oxnet PR where we'll have a `VlanId` newtype to share this across repositories. I've punted on a newtype for the moment (oxnet PR incoming in a day or two). This tries to make things consistent across the board.

### DIFF
--- a/common/src/network.rs
+++ b/common/src/network.rs
@@ -41,14 +41,31 @@ pub fn generate_ipv6_link_local(mac: MacAddr) -> Ipv6Addr {
 #[derive(Error, Debug, Clone)]
 pub enum VlanError {
     /// Not a valid VLAN ID
-    #[error("Invalid VLAN tag: {}", .0)]
+    #[error("Invalid VLAN ID {}, must be in the range 1..=4094", .0)]
     InvalidVlan(u16),
 }
 
+/// Validate a VLAN ID against the configurable range `1..=4094`.
+///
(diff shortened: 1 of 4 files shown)


zeeshanlakhani:
As @taspelund, he's started on https://github.com/oxidecomputer/oxnet/pull/85 and I'm working on a oxnet PR off his work there. This is a stop-gap until the oxnet work lands is updated throughout our ecosystem.

taspelund (approved):
Changes look good. I'm excited to replace this with an oxnet type, but I know it's coming :-)

zeeshanlakhani:
> Changes look good. I'm excited to replace this with an oxnet type, but I know it's coming :-)

for sure. I hope to have a full oxnet (based on your branch) piece up for tmrw.

rcgoodfellow:
I suspect there is a reason that Nils had constrained VIDs to be >= 2. @zeeshanlakhani @taspelund this PR will need e2e testing on a racklette with VID=1 in play.

PR #358, opened 2026-08-31
Ensure tfportd and dpd use unique Oximeter producer IDs

### LINKED ISSUE #357: `dpd` and `tfportd` use the same UUID for their Oximeter producer servers
This is fun. We construct a "UUID" from a few of the switch-specific identifiers in `dpd` here:

https://github.com/oxidecomputer/dendrite/blob/0fceb6bd4ad5ec7b7593d8c973f8d0e58d590616/dpd/src/oxstats.rs#L659-L665

Then do the exact same thing in `tfportd` here:

https://github.com/oxidecomputer/dendrite/blob/0fceb6bd4ad5ec7b7593d8c973f8d0e58d590616/tfportd/src/oxstats.rs#L146-L152

Obviously, this is bad. Oximeter is flapping between polling these two producers, depending on who registered last.

bnaecker:
Fixes #357

### DIFF
--- a/tfportd/src/oxstats.rs
+++ b/tfportd/src/oxstats.rs
@@ -58,6 +58,20 @@ pub(crate) struct OximeterConfig {
     switch_identifiers: types::SwitchIdentifiers,
 }
 
+impl OximeterConfig {
+    /// Generate our producer UUID based on the switch identifiers.
+    pub(crate) const fn producer_id(&self) -> Uuid {
+        // NOTE: `dpd` uses only the XOR of these three fields to construct
+        // _its_ producer ID. We can't do the same thing, otherwise we'll
+        // conflict and Oximeter will flap between polling `dpd` and us. Flip
+        // the last bit as a stupid way to ensure uniqueness.
+        let x = self.sled_identifiers.rack_id.as_u128()
+            ^ self.sled_identifiers.sled_id.as_u128()
+            ^ self.switch_identifiers.sidecar_id.as_u128();
+        Uuid::from_u128(x ^ 0b1)
+    }
+}
+
 /// Link tracker for tracking and untracking link names set by the program.
 pub struct LinkTracker {
     tracked_links: RwLock<HashSet<(String, ModelType)>>,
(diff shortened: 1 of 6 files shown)


tfportd/src/oxstats.rs
```diff
@@ -58,6 +58,20 @@ pub(crate) struct OximeterConfig {
...
+    /// Generate our producer UUID based on the switch identifiers.
+    pub(crate) const fn producer_id(&self) -> Uuid {
+        // NOTE: `dpd` uses only the XOR of these three fields to construct
+        // _its_ producer ID. We can't do the same thing, otherwise we'll
+        // conflict and Oximeter will flap between polling `dpd` and us. Flip
+        // the last bit as a stupid way to ensure uniqueness.
+        let x = self.sled_identifiers.rack_id.as_u128()
+            ^ self.sled_identifiers.sled_id.as_u128()
+            ^ self.switch_identifiers.sidecar_id.as_u128();
+        Uuid::from_u128(x ^ 0b1)
```
jgallagher, line 71: Could/should we put `OximeterConfig` in a crate shared between dpd/tfportd, and give it separate methods for `dpd_producer_id()` / `tfportd_producer_id()`? A separate crate is a little annoying, but that would (a) colocate these so they're more obviously correct and (b) make it more likely if someone wants to add stats to another producer in this repo, they can follow the existing pattern. (What does mgd do for oxstats?)

bnaecker: I didn't think it was worth it, but I could be convinced.

> What does mgd do for oxstats?

It uses a random UUID, but it's in a separate repo anyway..

bnaecker: So an issue here is that for some reason we can't figure out, the `OximeterConfig` type is _versioned_. It appears in the `OximeterMetadata` type that's returned from `oximeter_collect_meta_endpoint()` in the `dpd` API trait. It has no business being there, since it's entirely an implementation detail of `dpd`. I'm going to file an issue to remove that completely, and then we can put this configuration type into its own crate to make this work a little better.

jgallagher: Chatted about this offline briefly; there's quite a lot of confusing stuff going on here (including an `unpublished` endpoint) that is worth fixing but not as a blocker to this change.

jgallagher (approved):

tfportd/src/oxstats.rs
bnaecker: Filed https://github.com/oxidecomputer/dendrite/issues/359 to track

PR #361, opened 2026-09-01
[mcast, on] default multicast feature across all crates

zeeshanlakhani:
This enables the multicast feature on by default across crates and adds `--no-multicast` to exclude mcast from the shipped P4 codegen build.

With the feature on by default, the packet test harness no longer needs its MULTICAST conditional or any redundant setup.


cfzimmerman (commented):
Is it possible to remove the multicast feature flag altogether? Or is there a path to doing so? Removing the cargo feature would be a giant win imo.

If compiling and running with no-multicast is still a requirement, I feel like we should still be testing that config in CI.

zeeshanlakhani:
> Is it possible to remove the multicast feature flag altogether? Or is there a path to doing so? Removing the cargo feature would be a giant win imo.
> 
> If compiling and running with no-multicast is still a requirement, I feel like we should still be testing that config in CI.

Yeah, this was my original intention (removing it altogether). I guess, let's do that now (going back to something @rmustacc mentioned in a sync awhile back).

zeeshanlakhani:
Closing in favor of https://github.com/oxidecomputer/dendrite/pull/371.

PR #362, opened 2026-09-01
[p4, sidecar] fix underlay/external mcast counter ops

zeeshanlakhani:
A replica still carrying a Geneve header at the egress counter block is bound for an underlay port, while external replicas are decapsulated earlier, at ingress, for external-only groups or in `mcast_egress` for bifurcated (external + underlay members) ones. Previously, we used a conditional:

```p4
else if (hdr.geneve.isValid()) {
```

... on Geneve hdr validity too early in the chain to determine an external- vs underlay-destined packet, and we didn't consider using the `is_egress_rid_mcast` check at all.

## The Fix

Now, we treat the three conditions for link-local, external mcast, and underlay mcast independently and more explicitly given the egress rid, which is 0 for link-local multicast packets and combines with Geneve hdr validity to determine the count for underlay vs external.

The underlay and external arms key on `is_egress_rid_mcast` and not the negation of `is_link_local_ipv6_mcast`: the compiler doesn't carry the negation into the later gateways of the pipeline, which meant that a link-local packet incremented `external_mcast_ctr` alongside its own counter. This is now handled because link-local replicas arrive with `egress_rid` of 0, and scope 2 is refused at group creation, so no replicated packet ever carries a link-local outer destination.

### DIFF
--- a/dpd/p4/sidecar.p4
+++ b/dpd/p4/sidecar.p4
@@ -2346,29 +2346,33 @@ control Egress(
 		} else if (is_mcast == true) {
 			mcast_ctr.count(eg_intr_md.egress_port);
 
-			// Count each multicast replica by type, one simple gateway
-			// per condition:
+			// Per-type replica counters. There are three disjoint
+			// cases presented here, with one simple gateway each.
+			// The compiler cannot carry a negated condition across
+			// gateways, so each case gets its own gateway instead
+			// of an else branch (the IngressDeparser notes the same
+			// limitation):
 			//
-			//  - link-local (ff02::/16 outer dst) — never replicated by
-			//    the PRE; it arrives with egress_rid == 0.
-			//  - egress_rid > 0, Geneve hdr still valid — bound for an
-			//    underlay port.
-			//  - egress_rid > 0, no Geneve hdr — decapped at ingress
-			//    (external-only groups) or by the mcast_egress control
-			//    above (bifurcated, external + underlay members), and
-			//    bound for an external member.
+			// - link-local (ff02::/16 outer dst): always forwarded,
+			//   never PRE (Packet Replication Engine)-replicated;
+			//   it arrives with egress_rid == 0.
+			// - egress_rid > 0 + a valid Geneve hdr: an underlay
+			//   replica that's still encapsulated.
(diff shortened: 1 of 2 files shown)


dpd/p4/sidecar.p4
```diff
@@ -2346,15 +2346,38 @@ control Egress(
...
+			// for the underlay still carries its encapsulation whether
+			// or not the group is bifurcated.
+			//
+			// The rid arms don't test !is_link_local_ipv6_mcast: the
+			// compiler doesn't carry the negation across gateways.
+			// They don't need to either, since scope 2 is refused at
+			// group creation, so a replica never carries a link-local
+			// outer destination.
+			//
+			// The IngressDeparser above notes the same compiler limitation.
```
cfzimmerman: I have a hard time groking this comment. Is it possible to digest it a bit more?

cfzimmerman (commented):
Looks reasonable, but again I'd like to defer final approval to someone with more knowledge of the sidecar program.

dpd/p4/sidecar.p4
```diff
@@ -2346,15 +2346,42 @@ control Egress(
 		} else if (is_mcast == true) {
 			mcast_ctr.count(eg_intr_md.egress_port);
 
+			// Per-type replica counters. There are three disjoint
+			// cases presented here, with one simple gateway each.
```
FelixMcFelix, line 2350: ...Gateway?

dpd/p4/sidecar.p4
```diff
@@ -2346,15 +2346,42 @@ control Egress(
...
+			// The compiler cannot carry a negated condition across
+			// gateways, so each case gets its own gateway instead
+			// of an else branch (the IngressDeparser notes the same
+			// limitation):
```
FelixMcFelix, line 2354: `IngressDeparser` has the issue that writing it more cleanly uses too many checksum engines, because the compiler doesn't recognise the disjoint cases for some reason. Do we end up running out of _stages_ if we do a sensible `if (link_local) { /* */ } else if (replicated) {if (geneve) { /* */ } else { /* */ }}` here, or is the list of `if()`s defensive? I.e. the existing conditions were broken, but we didn't hit a resource limit from using else-ifs AFAICT.

PR #364, opened 2026-09-02
[dpd, mcast] Tighten multicast group and source validation

zeeshanlakhani:
This work constrains group and SSM source validation based on a variety of specs, linked throughout the PR. Previously, DPD was accepting groups and (S, G) sources that no packet could actually exercise, including ones that can never leave a host/link or, worse, sources a switch can't forward.

Here's the list (the first classifier one was the most obvious):

| address                        | verdict | spec                          |
|--------------------------------|---------|-------------------------------|
| ff3x::/32 in ff30::/12 only    | is SSM  | RFC 4607 §1; RFC 3306 is ASM  |
| ff3x:: group ID > 0x40000000   | accept  | RFC 4607 §4.3                 |
| v6 scopes other than 4/5/8/e   | reject  | RFC 7346; no Ethernet realm   |
| class E (240.0.0.0/4) source   | reject  | RFC 6890 "Source: False"      |
| 0.0.0.0/8 source               | reject  | RFC 6890 "Forwardable: False" |
| 100.64.0.0/10 source           | accept  | RFC 6598, operator-internal   |
| source/group family mismatch   | reject  | (local choice)                |
| 232.0.0.0 (SSM null)           | reject  | RFC 4607 §4.3                 |
| v4-mapped/compat v6 source     | reject  | RFC 4291 §2.5.5, aliases IPv4 |

### DIFF
--- a/dpd/src/mcast/validate.rs
+++ b/dpd/src/mcast/validate.rs
@@ -17,10 +17,11 @@ use common::network::NatTarget;
 use dpd_types::mcast::MulticastTag;
 use omicron_common::address::{
     IPV4_LINK_LOCAL_MULTICAST_SUBNET, IPV4_SSM_SUBNET,
-    IPV6_INTERFACE_LOCAL_MULTICAST_SUBNET, IPV6_LINK_LOCAL_MULTICAST_SUBNET,
-    IPV6_RESERVED_SCOPE_MULTICAST_SUBNET, IPV6_SSM_SUBNET,
     UNDERLAY_MULTICAST_SUBNET,
 };
+
+const IPV4_SSM_RESERVED_NULL: Ipv4Addr = Ipv4Addr::new(232, 0, 0, 0);
+const IPV6_SSM_MIN_VALID_GROUP_ID: u32 = 0x4000_0001;
 
 /// Check if an IP address is unicast (emulating the unstable std::net API).
 /// For IP addresses, unicast means simply "not multicast".
@@ -35,6 +36,8 @@ pub(crate) fn validate_multicast_address(
     addr: IpAddr,
     sources: Option<&[IpSrc]>,
 ) -> DpdResult<()> {
+    validate_source_address_family(addr, sources)?;
+
     // First validate that source addresses are unicast
     validate_source_addresses(sources)?;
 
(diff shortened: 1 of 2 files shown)


dpd/src/mcast/validate.rs
```diff
@@ -237,13 +302,30 @@ fn validate_ipv4_source_address(ipv4: Ipv4Addr) -> DpdResult<()> {
...
         )));
     }
+
+    let embedded = match ipv6.segments() {
+        [0, 0, 0, 0, 0, 0xffff, ..] => {
+            Some("IPv4-mapped (::ffff:0:0/96, RFC 4291 §2.5.5.2)")
+        }
+        [0, 0, 0, 0, 0, 0, ..] => {
+            Some("IPv4-compatible (::/96, RFC 4291 §2.5.5.1)")
+        }
```
cfzimmerman, line 323: cool, TIL this exists. Thanks for the RFC references.

dpd/src/mcast/validate.rs
```diff
@@ -131,8 +173,40 @@ fn validate_ipv6_multicast(
         )));
     }
 
+    // Allow admin-local (4), site-local (5), organization-local (8), and
+    // global (e) scopes. Reject reserved, interface-local, link-local,
+    // realm-local, and unassigned scopes. See RFC 7346 §2.
+    let scope = addr.segments()[0] & 0x000f;
+    let scope_name = match scope {
```
cfzimmerman: I got a bit tripped up when first reading this. What do you think about renaming `scope_name` to something like `unsupported_scope`? Or any name that indicates that `Some` is actually something we want to reject.

dpd/src/mcast/validate.rs
```diff
@@ -69,10 +94,21 @@ pub(crate) fn validate_nat_target(nat_target: NatTarget) -> DpdResult<()> {
...
+/// [RFC 4607 §1]: https://www.rfc-editor.org/rfc/rfc4607#section-1
+/// [RFC 3306]: https://www.rfc-editor.org/rfc/rfc3306
 pub(crate) fn is_ssm(addr: IpAddr) -> bool {
     match addr {
         IpAddr::V4(ipv4) => IPV4_SSM_SUBNET.contains(ipv4),
-        IpAddr::V6(ipv6) => IPV6_SSM_SUBNET.contains(ipv6),
+        IpAddr::V6(ipv6) => {
+            let segs = ipv6.segments();
+            segs[0] & 0xfff0 == 0xff30 && segs[1] == 0
+        }
```
cfzimmerman, line 115: Claude's review on this PR mostly questioned divergences and duplication between this module and its counterpart in omicron. For example, ipv6 ssm validation:

https://github.com/oxidecomputer/omicron/blob/9d95e0cf4542d2bc776c40b69f97ba6e05cff199/common/src/address.rs#L68-L69

https://github.com/oxidecomputer/omicron/blob/9d95e0cf4542d2bc776c40b69f97ba6e05cff199/nexus/src/app/ip_pool.rs#L85-L87

Is there a future where we can have a single source of truth? Or do you have a plan for keeping the two in sync?

zeeshanlakhani: There's a goal to move things to oxnet. I don't want to conflate that right now, hence the types in the repos. But, it's the plan and we can do it the changes once when that is over the line.

PR #365, opened 2026-09-02
[dpd-api] correct external multicast group scope docs

zeeshanlakhani:
The external-group docs say “IPv4 and non-admin-local IPv6 multicast traffic,” but external groups can use admin-local IPv6 too. The exception, however, is the `ff04::/64` underlay range, which is reserved for the underlay mcast group mapped from the external mcast group. This makes the information explicit.

The old wording is in a blessed spec, so this needs a new API version: `MCAST_EXTERNAL_SCOPE_DOCS`.

### DIFF
--- a/dpd-api/src/lib.rs
+++ b/dpd-api/src/lib.rs
@@ -9,7 +9,7 @@
 use std::collections::BTreeMap;
 use std::net::{Ipv4Addr, Ipv6Addr};
 
-use dpd_types_versions::{latest, v1, v4, v7};
+use dpd_types_versions::{latest, v1, v4, v7, v8};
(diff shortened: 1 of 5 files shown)


cfzimmerman (commented):
Could these documented constraints also be made self-enforcing via newtypes? Your decision whether doing so is worth the API churn and code complexity, but imo it's worth considering.

zeeshanlakhani:
> Could these documented constraints also be made self-enforcing via newtypes? Your decision whether doing so is worth the API churn and code complexity, but imo it's worth considering.

I did consider this. I will push that update shortly.

PR #366, opened 2026-09-02
[swadm,multicast] add multicast inspection subcommand (->list, ->get)

zeeshanlakhani:
This adds a READ-only (no-mutation) `swadm multicast` subcommand (aliased as `mcast` as well) for inspecting the multicast groups programmed on the switch.

The two commands are essentially a LIST and GET:

- `multicast list [-t <tag>] [-k external|underlay]` renders all groups as a table view, filtered by tag, by replication kind, or both.
- `multicast get <group-ip>` shows one group's full configuration.

The kind filter matches on the response type: `external` means an external group with a NAT target and no (underlying) direction membership; `underlay` groups sit in the reserved ff04::/64 subnet and replicate to the member ports (to guest subscribers).

### DIFF
--- a/swadm/src/multicast.rs
+++ b/swadm/src/multicast.rs
@@ -1,0 +1,333 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/
+//
+// Copyright 2026 Oxide Computer Company
+
+//! Inspection of the multicast groups programmed on the switch.
+//!
+//! Here's an example set of output against a running dpd bin:
+//!
+//! ```text
+//! $ swadm multicast list
+//! GROUP IP       KIND      EXT GROUP ID  UL GROUP ID  TAG         DETAIL
+//! 224.0.1.50     external  65532         -            oxide-demo  nat=ff04::2 mac=33:33:00:00:00:02 vni=88 vlan=- src=any
+//! 232.123.45.99  external  65534         -            oxide-demo  nat=ff04::1 mac=33:33:00:00:00:01 vni=77 vlan=10 src=10.0.0.1,10.0.0.2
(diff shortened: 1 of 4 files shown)


swadm/tests/multicast.rs
cfzimmerman, line 1: `swadm` now has common infra for tests. Can we move this module over to the new expectorate pattern?

https://github.com/oxidecomputer/dendrite/pull/350

swadm/src/multicast.rs
```diff
@@ -0,0 +1,333 @@
...
+
+    match group {
+        types::MulticastGroupResponse::Underlay {
+            group_ip,
+            external_group_id,
+            underlay_group_id,
+            tag,
+            members,
+        } => {
+            println!("Group IP:          {group_ip}");
```
cfzimmerman, line 272: Can we use `tabwriter` here instead of manual spacing?

swadm/tests/multicast.rs
cfzimmerman: In my understanding, this test uses the dropshot client to create/delete multicast assets so that we can then inspect the output using swadm. Doing so requires the updated tokio dependency.

First, I'd like to use the shared expectorate infra for swadm validation as stated above.

Taking it further, can we add multicast asset creation/deletion to swadm so that this test doesn't need the dropshot client at all?

Doing so creeps the scope of this PR, and restricting to dropshot makes sense if you know of a reason swadm definitely should not support multicast creation/deletion. But adding that to swadm sounds like useful functionality and would simplify this test.

Simplifying swadm tests isn't an end goal in itself, but swadm tests are close enough to human/hand testing that I want to be thoughtful. For example, could you then set up a multicast demo just using cli commands? That would be pretty cool.

cfzimmerman (changes requested):

swadm/tests/multicast.rs
cfzimmerman: Btw the "request changes" is for you to please check out the new expectorate test structure. Which was merged since you opened this PR, so apologies for the churn!
Adding create/delete and using tabwriter in `multicast_get` are your decision.

PR #370, opened 2026-09-03
Respect address tags in port_settings_*

### LINKED ISSUE #342: port_settings_apply clobbers link-local addresses added via link_ipv6_create
When `port_settings_apply` is run, it grabs a copy of dpd's internal state for each port/link, diffs it against the state requested by the API client, and then reconciles the difference.

The problem we have here is that omicron is the main (practically only) consumer of `port_settings_apply`, and it doesn't include any link-local IPv6 addresses in its request... since these are handled by tfportd and illumos.

tfportd learns about IPv6 link-local addresses in illumos from its reconciler, and synchronizes them into dpd via `link_ipv6_create`.

Both `port_settings_apply` and `link_ipv6_create` modify the same internal state: `Link.ipv6`
```
#[derive(Debug)]
pub struct Link {
    pub port_id: PortId,
    pub link_id: LinkId,
    pub port_hdl: PortHdl,
    pub asic_port_id: AsicId,
    pub updated: i64,
    pub presence: bool,
    pub ipv6_enabled: bool,
    pub tx_eq: Option<TxEq>,
    pub fsm_state: asic::PortFsmState,
    pub link_state: LinkState,
    pub media: PortMedia,

    /// A list of IPv4 addresses assigned to this link.
    pub ipv4: BTreeSet<Ipv4Entry>,

    /// A list of IPv6 addresses assigned to this link.
    pub ipv6: BTreeSet<Ipv6Entry>,                    <<<<<<<<<<<

    pub linkup_tracker: LinkUpTracker,
    pub autoneg_tracker: AutonegTracker,

    /// Configuration requested by the user/sled-agent.
    pub config: LinkConfig,

    /// State that actually exists in the ASIC layer.
    pub(crate) plumbed: LinkPlumbed,
}
```

When `port_settings_apply` is run, the internal dpd state includes all items in `Link.ipv6` (including the link-locals added by tfportd) and thus the diff always indicates that the link-local is old state that should be removed.

This results in the omicron reconciler (via `port_settings_apply`) and the tfport reconciler (via `link_ipv6_create`) battling each other and the IPv6 link-local addresses being removed from the ASIC anytime omicron's reconciler runs after the tfport reconciler... which means that periodically, the ASIC will stop punting traffic destined for that link-local address. This would affect NDP, BGP unnumbered, etc. and could potentially cause service disruptions if the blackholing were to last long enough.

cfzimmerman:
Reproduces and fixes https://github.com/oxidecomputer/dendrite/issues/342
See that issue if you have questions about this design.

### Tests

These tests failed before the second commit.

```txt
failures:
    chaos_tests::port_settings::address_cmds_respect_tags
    chaos_tests::port_settings::apply_fails_on_tag_conflict
    chaos_tests::port_settings::settings_apply_respects_tags
    chaos_tests::port_settings::settings_clear_ignores_tags
```

- `settings_apply_respects_tags` is the most important. It exercises the basic enforcement of tag separation. One tag creates an address. Another tag applies an address. The first tag's address still needs to exist. This is the motivating issue between tfportd and sled-agent. This test fails when tag1 lists addresses and sees that its address is missing.
- `address_cmds_respect_tags` verifies tags are respected by address commands of the type used in swadm, tfportd, and maghemite. CRUD itself works, but the test fails in verification when `port_settings_get` returns more addresses than `link_ipv*_list`.
- `apply_fails_on_tag_conflict` is a weird but meaningful edge case. The passing test `create_addr_rejects_overwrites` demonstrates that creating an address fails if the address already exists (under another tag). `port_settings_apply` should allow re-declaring addresses of the same tag, but it should fail when creating an address that belongs to a different tag rather than stealing the address from that tag. Stealing the address complicates deletion, which is implicated if a transaction ever needs to be rolled back. This particular test fails on "Addresses belong to another tag" because stealing is allowed in the current implementation.
- `settings_clear_ignores_tags` exercises the same issue as `settings_apply_respects_tags`. If I remove the `tag1` check, then this passes. So it's more collateral damage than a legit failure. After all, the test is meant to describe the current behavior rather than declare what it should be.

### Summary

The main change is in `port_settings.rs`, where we merge addresses belonging to other tags into the diffable representation.

### Context

This PR makes a decision about address tagging in DPD, and it tries to avoid other structural changes. This in-progress stacked PR is exploring solutions for txn test failures and ip addr code simplification: https://github.com/oxidecomputer/dendrite/pull/374

### Decisions

- Tagged addresses are stored in a `BTreeMap<Ipv*Addr, String>`. So an address may necessarily only be added to a link by a single tag. I believe this was the previous behavior too, but it's more self-enforcing than `BTreeSet<Ipv*Entry>`. I would like a more descriptive `Tag` type, but that implies a lot of API changes and felt out of scope for this PR. Right now we're just trying to make the current API work as expected.
- Currently addresses are the only tagged resource on a link.
- If a resource is tagged, it is unconditionally tagged. This was already the case in the `Link` struct, but that invariant was a bit hidden. Where the dropshot api uses `Option<String>` as a tag, we unwrap that to `""` in `api_server.rs`. We were already doing this in most/all places, but now it's more visible at the top layer.
- `LinkSpec` holds addresses for all tags, which is necessary to avoid dropping other tags during `port_settings_apply`. This contrasts with `LinkSettings`, which only reflects addresses from a single tag.

# Validation

This PR warrants system-level testing. I'm currently setting up an eval env, and I intend to post a voxel smoke test soon.

### DIFF
--- a/dpd-client/tests/chaos_tests/port_settings.rs
+++ b/dpd-client/tests/chaos_tests/port_settings.rs
@@ -9,18 +9,24 @@ use super::harness::{
     new_dpd_client, run_dpd,
 };
 use super::util::{link_list_ipv4, link_list_ipv6};
+use crate::chaos_tests::harness;
+use crate::chaos_tests::util::HttpResponseCheck;
+use crate::chaos_tests::util::IpRng;
+
+use anyhow::bail;
 use asic::chaos::{AsicConfig, Chaos, TableChaos};
 use asic::table_chaos;
 use common::table::TableType;
 use dpd_client::types::{
-    LinkCreate, LinkId, LinkSettings, PortFec, PortId, PortSettings, PortSpeed,
+    Ipv4Entry, Ipv6Entry, LinkCreate, LinkId, LinkSettings, PortFec, PortId,
+    PortSettings, PortSpeed,
 };
 use dpd_client::{Client, ROLLBACK_FAILURE_ERROR_CODE};
 use http::status::StatusCode;
 use pretty_assertions::{Comparison, assert_eq};
 use rand::Rng;
 use std::collections::HashMap;
-use std::net::{Ipv4Addr, Ipv6Addr};
+use std::net::{IpAddr, Ipv4Addr, Ipv6Addr};
 use std::sync::Arc;
 use std::sync::atomic::{AtomicU8, Ordering};
 use tokio::time::Duration;
@@ -31,6 +37,17 @@ const TESTING_RADIX: usize = 33;
 // we give up?
 const RETRY_INTERVAL: Duration = Duration::from_millis(200);
 const RETRY_MAX: Duration = Duration::from_secs(5);
+
+/// A `LinkCreate` config with common defaults.
+const LINK_CREATE: LinkCreate = LinkCreate {
+    lane: None,
+    autoneg: false,
+    kr: false,
+    speed: PortSpeed::Speed100G,
+    fec: Some(PortFec::None),
+    tx_eq: None,
+    allow_ddm_traffic: false,
+};
 
 #[cfg(test)]
 mod retry {
@@ -83,18 +100,7 @@ async fn test_basic_autoneg_chaos() -> anyhow::Result<()> {
     let (_guard, client) = init_harness("autoneg", &config);
 
     let err = client
-        .link_create(
-            &"qsfp0".parse().unwrap(),
-            &LinkCreate {
-                lane: None,
-                autoneg: false,
-                kr: false,
-                speed: PortSpeed::Speed100G,
-                fec: Some(PortFec::None),
-                tx_eq: None,
-                allow_ddm_traffic: false,
-            },
-        )
+        .link_create(&"qsfp0".parse().unwrap(), &LINK_CREATE)
         .await
         .expect_err("Expected error on create");
 
@@ -122,15 +128,7 @@ async fn test_port_settings_addr_fail_1() -> anyhow::Result<()> {
     settings.links.insert(
         "0".into(),
         LinkSettings {
-            params: LinkCreate {
-                lane: None,
-                autoneg: false,
-                kr: false,
-                fec: Some(PortFec::None),
-                speed: PortSpeed::Speed100G,
-                tx_eq: None,
-                allow_ddm_traffic: false,
-            },
+            params: LINK_CREATE,
             addrs: vec!["203.0.113.47".parse().unwrap()],
         },
     );
@@ -164,15 +162,7 @@ async fn test_port_settings_addr_success_1() -> anyhow::Result<()> {
     settings.links.insert(
         "0".into(),
         LinkSettings {
-            params: LinkCreate {
-                lane: None,
-                autoneg: false,
-                kr: true,
-                fec: Some(PortFec::None),
-                speed: PortSpeed::Speed100G,
-                tx_eq: None,
-                allow_ddm_traffic: false,
-            },
+            params: LinkCreate { kr: true, ..LINK_CREATE },
             addrs: vec!["203.0.113.47".parse().unwrap()],
         },
     );
@@ -204,15 +194,7 @@ async fn test_port_settings_addr_success_multi() -> anyhow::Result<()> {
     settings.links.insert(
         "0".into(),
         LinkSettings {
-            params: LinkCreate {
-                lane: None,
-                autoneg: false,
-                kr: true,
-                fec: Some(PortFec::None),
-                speed: PortSpeed::Speed100G,
-                tx_eq: None,
-                allow_ddm_traffic: false,
-            },
+            params: LinkCreate { kr: true, ..LINK_CREATE },
             addrs: vec!["203.0.113.47".parse().unwrap()],
         },
     );
@@ -234,15 +216,7 @@ async fn test_port_settings_addr_success_multi() -> anyhow::Result<()> {
     settings.links.insert(
         "0".into(),
         LinkSettings {
-            params: LinkCreate {
-                lane: None,
-                autoneg: false,
-                kr: true,
-                fec: Some(PortFec::None),
-                speed: PortSpeed::Speed100G,
-                tx_eq: None,
-                allow_ddm_traffic: false,
-            },
+            params: LinkCreate { kr: true, ..LINK_CREATE },
             addrs: vec![
                 "203.0.113.46".parse().unwrap(),
                 "203.0.113.48".parse().unwrap(),
@@ -275,15 +249,7 @@ async fn test_port_settings_addr_success_multi() -> anyhow::Result<()> {
     settings.links.insert(
         "0".into(),
         LinkSettings {
-            params: LinkCreate {
-                lane: None,
-                autoneg: false,
-                kr: true,
-                fec: Some(PortFec::None),
-                speed: PortSpeed::Speed100G,
-                tx_eq: None,
-                allow_ddm_traffic: false,
-            },
+            params: LinkCreate { kr: true, ..LINK_CREATE },
             addrs: vec![
                 "203.0.113.47".parse().unwrap(),
                 "fd00:1701::d".parse().unwrap(),
@@ -551,4 +517,528 @@ fn random_port_settings() -> PortSettings {
             LinkSettings { params, addrs },
         )]),
     }
+}
+
+const TAG1: &str = "chaos1";
+const TAG2: &str = "chaos2";
+
+/// Verifies tagged port_settings_apply actions don't affect
+/// resources from other tags.
+#[tokio::test]
+async fn settings_apply_respects_tags() -> anyhow::Result<()> {
+    let no_failures = AsicConfig::uniform_set(TESTING_RADIX, 0.);
+    let (_guard, client) =
+        harness::init_harness("settings_apply_respects_tags", &no_failures);
+
+    let mut rng = IpRng::new(12345);
+    let port_id: PortId = "qsfp0".parse()?;
+    let link_id =
+        client.link_create(&port_id, &LINK_CREATE).await?.into_inner();
+
+    let tag1 = TestAddrs::new(
+        &mut rng,
+        TAG1.to_string(),
+        &client,
+        port_id.clone(),
+        link_id,
+    );
+    let tag2 = TestAddrs::new(
+        &mut rng,
+        TAG2.to_string(),
+        &client,
+        port_id.clone(),
+        link_id,
+    );
+
+    tag1.create_addrs().await?;
+    tag2.apply_addrs().await?;
+
+    tag1.verify_addrs_exist(Verify::NonExhaustive).await?;
+    tag2.verify_addrs_exist(Verify::NonExhaustive).await?;
+
+    client
+        .port_settings_apply(
+            &port_id,
+            Some(TAG2),
+            &PortSettings {
+                links: HashMap::from([(
+                    link_id.to_string(),
+                    LinkSettings { params: LINK_CREATE, addrs: Vec::new() },
+                )]),
+            },
+        )
+        .await?;
+
+    tag1.verify_addrs_exist(Verify::Exhaustive).await?;
+
+    Ok(())
+}
+
+/// Verifies tagged address_*_create and delete don't affect
+/// resources under different tags.
+#[tokio::test]
+async fn address_cmds_respect_tags() -> anyhow::Result<()> {
+    let no_failures = AsicConfig::uniform_set(TESTING_RADIX, 0.);
+    let (_guard, client) =
+        harness::init_harness("address_cmds_respect_tags", &no_failures);
+
+    let mut rng = IpRng::new(54321);
+    let port_id: PortId = "qsfp0".parse()?;
+    let link_id = LinkId(0);
+
+    let tag1 = TestAddrs::new(
+        &mut rng,
+        TAG1.to_string(),
+        &client,
+        port_id.clone(),
+        link_id,
+    );
+    let tag2 = TestAddrs::new(
+        &mut rng,
+        TAG2.to_string(),
+        &client,
+        port_id.clone(),
+        link_id,
+    );
+
+    tag2.apply_addrs().await?;
+
+    client
+        .link_ipv4_create(
+            &port_id,
+            &link_id,
+            &Ipv4Entry { addr: tag2.v4_entry.addr, tag: TAG1.to_string() },
+        )
+        .await
+        .expect_err(
+            "Registering the same address under different tags should fail",
+        )
+        .expect_status(StatusCode::CONFLICT);
+
+    tag1.create_addrs().await?;
+
+    tag1.verify_addrs_exist(Verify::NonExhaustive).await?;
+    tag2.verify_addrs_exist(Verify::NonExhaustive).await?;
+
+    client.link_ipv4_delete(&port_id, &link_id, &tag1.v4_entry.addr).await?;
+    client.link_ipv6_delete(&port_id, &link_id, &tag1.v6_entry.addr).await?;
+
+    tag2.verify_addrs_exist(Verify::Exhaustive).await?;
+    tag1.verify_addrs_exist(Verify::NonExhaustive)
+        .await
+        .expect_err("tag1 addresses should have been deleted");
+
+    Ok(())
+}
+
+/// Verifies the current known behavior of port_settings_clear, which
+/// is to delete the link and all its config from the port.
+///
+/// This test isn't necessarily endorsing the implementation, but
+/// it does seek to track API's current behavior.
+#[tokio::test]
+async fn settings_clear_ignores_tags() -> anyhow::Result<()> {
+    let no_failures = AsicConfig::uniform_set(TESTING_RADIX, 0.);
+    let (_guard, client) =
+        harness::init_harness("settings_clear_ignores_tags", &no_failures);
+
+    let mut rng = IpRng::new(1010101);
+    let port_id: PortId = "qsfp0".parse()?;
+    let link_id =
+        client.link_create(&port_id, &LINK_CREATE).await?.into_inner();
+
+    let tag1 = TestAddrs::new(
+        &mut rng,
+        TAG1.to_string(),
+        &client,
+        port_id.clone(),
+        link_id,
+    );
+    let tag2 = TestAddrs::new(
+        &mut rng,
+        TAG2.to_string(),
+        &client,
+        port_id.clone(),
+        link_id,
+    );
+
+    tag1.create_addrs().await?;
+    tag2.apply_addrs().await?;
+
+    tag1.verify_addrs_exist(Verify::NonExhaustive).await?;
+    tag2.verify_addrs_exist(Verify::NonExhaustive).await?;
+
+    // Tag2 means nothing here
+    client.port_settings_clear(&port_id, Some(TAG2)).await?;
+
+    tag1.verify_addrs_exist(Verify::Exhaustive).await.expect_err(
+        "Addresses do not exist because we cleared the port settings.",
+    );
+    tag2.verify_addrs_exist(Verify::NonExhaustive).await.expect_err(
+        "Addresses do not exist because we cleared the port settings.",
+    );
+
+    Ok(())
+}
+
+/// Verifies that link address creation fails when that
+/// address has already been registered (under another tag).
+#[tokio::test]
+async fn create_addr_rejects_overwrites() -> anyhow::Result<()> {
+    let no_failures = AsicConfig::uniform_set(TESTING_RADIX, 0.);
+    let (_guard, client) =
+        harness::init_harness("create_addr_overwrite_rejected", &no_failures);
+    let mut rng = IpRng::new(1433);
+    let port_id: PortId = "qsfp0".parse()?;
+    let link_id =
+        client.link_create(&port_id, &LINK_CREATE).await?.into_inner();
+
+    let tag1 = TestAddrs::new(
+        &mut rng,
+        TAG1.to_string(),
+        &client,
+        port_id.clone(),
+        link_id,
+    );
+
+    tag1.apply_addrs().await?;
+
+    client
+        .link_ipv4_create(&port_id, &link_id, &tag1.v4_entry)
+        .await
+        .expect_err("Duplicate ipv4 registration fails")
+        .expect_status(StatusCode::CONFLICT);
+
+    client
+        .link_ipv6_create(&port_id, &link_id, &tag1.v6_entry)
+        .await
+        .expect_err("Duplicate ipv6 registration fails")
+        .expect_status(StatusCode::CONFLICT);
+
+    client
+        .link_ipv4_create(
+            &port_id,
+            &link_id,
+            &Ipv4Entry { addr: tag1.v4_entry.addr, tag: TAG2.to_string() },
+        )
+        .await
+        .expect_err("Cross-tag ipv4 registration fails")
+        .expect_status(StatusCode::CONFLICT);
+
+    client
+        .link_ipv6_create(
+            &port_id,
+            &link_id,
+            &Ipv6Entry { addr: tag1.v6_entry.addr, tag: TAG2.to_string() },
+        )
+        .await
+        .expect_err("Cross-tag ipv6 registration fails")
+        .expect_status(StatusCode::CONFLICT);
+
+    // The idempotent API necessarily accepts duplicate registrations.
+    tag1.apply_addrs().await?;
+
+    Ok(())
+}
+
+/// Verifies deletion fails when the address does not exist.
+#[tokio::test]
+async fn deleted_address_must_exist() -> anyhow::Result<()> {
+    let no_failures = AsicConfig::uniform_set(TESTING_RADIX, 0.);
+    let (_guard, client) =
+        harness::init_harness("delete_addr_conflicts", &no_failures);
+    let mut rng = IpRng::new(1516);
+    let port_id: PortId = "qsfp0".parse()?;
+    let link_id = LinkId(0);
+
+    let tag1 = TestAddrs::new(
+        &mut rng,
+        TAG1.to_string(),
+        &client,
+        port_id.clone(),
+        link_id,
+    );
+
+    tag1.apply_addrs().await?;
+
+    client
+        .link_ipv4_delete(&port_id, &link_id, &rng.unique_ipv4())
+        .await
+        .expect_err("Deleting a non-existent IPv4 addr fails")
+        .expect_status(StatusCode::NOT_FOUND);
+
+    client
+        .link_ipv6_delete(&port_id, &link_id, &rng.unique_ipv6())
+        .await
+        .expect_err("Deleting a non-existent IPv6 addr fails")
+        .expect_status(StatusCode::NOT_FOUND);
+
+    // Deletion is not currently tagged, so there's no way/reason to test
+    // tag conflicts.
+
+    client.link_ipv4_delete(&port_id, &link_id, &tag1.v4_entry.addr).await?;
+    client.link_ipv6_delete(&port_id, &link_id, &tag1.v6_entry.addr).await?;
+    tag1.verify_addrs_exist(Verify::NonExhaustive)
+        .await
+        .expect_err("Addresses were deleted");
+
+    Ok(())
+}
+
+/// One tagged `port_settings_apply` can delete a link created
+/// by another tag's `port_settings_apply`. This isn't functionality
+/// we necessarily want, but we should also never exercise it in practice.
+/// Since it's a subtle footgun, this test at least documents its existence
+/// and observes if this ever changes.
+#[tokio::test]
+async fn apply_implicit_delete() -> anyhow::Result<()> {
+    let no_failures = AsicConfig::uniform_set(TESTING_RADIX, 0.);
+    let (_guard, client) =
+        harness::init_harness("apply_implicit_delete", &no_failures);
+    let mut rng = IpRng::new(1548);
+    let port_id: PortId = "qsfp0".parse()?;
+
+    let tag1 = TestAddrs::new(
+        &mut rng,
+        TAG1.to_string(),
+        &client,
+        port_id.clone(),
+        LinkId(1),
+    );
+
+    let tag2 = TestAddrs::new(
+        &mut rng,
+        TAG2.to_string(),
+        &client,
+        port_id.clone(),
+        LinkId(2),
+    );
+
+    tag1.apply_addrs().await?;
+    tag1.verify_addrs_exist(Verify::Exhaustive).await?;
+
+    tag2.apply_addrs().await?;
+
+    tag1.verify_addrs_exist(Verify::NonExhaustive)
+        .await
+        .expect_err("Link 1 was deleted");
+    tag2.verify_addrs_exist(Verify::Exhaustive).await?;
+
+    Ok(())
+}
+
+/// port_settings_apply fails if it would overwrite an address
+/// belonging to another tag. Again, this should not arise in
+/// practice, but successful rollback is important in such cases.
+#[tokio::test]
+async fn apply_fails_on_tag_conflict() -> anyhow::Result<()> {
+    let no_failures = AsicConfig::uniform_set(TESTING_RADIX, 0.);
+    let (_guard, client) =
+        harness::init_harness("apply_tag_conflict", &no_failures);
+    let mut rng = IpRng::new(1616);
+    let port_id: PortId = "qsfp0".parse()?;
+    let link_id = LinkId(0);
+
+    let tag1 = TestAddrs::new(
+        &mut rng,
+        TAG1.to_string(),
+        &client,
+        port_id.clone(),
+        link_id,
+    );
+
+    client.link_create(&port_id, &LINK_CREATE).await?;
+    client
+        .link_ipv4_create(
+            &port_id,
+            &link_id,
+            &Ipv4Entry { addr: tag1.v4_entry.addr, tag: TAG2.to_string() },
+        )
+        .await?;
+    client
+        .link_ipv6_create(
+            &port_id,
+            &link_id,
+            &Ipv6Entry { addr: tag1.v6_entry.addr, tag: TAG2.to_string() },
+        )
+        .await?;
+
+    tag1.apply_addrs()
+        .await
+        .expect_err("Addresses belong to another tag")
+        .expect_status(StatusCode::CONFLICT);
+
+    client.link_ipv4_delete(&port_id, &link_id, &tag1.v4_entry.addr).await?;
+
+    tag1.apply_addrs()
+        .await
+        .expect_err("IPv6 address belongs to another tag")
+        .expect_status(StatusCode::CONFLICT);
+
+    client.link_ipv6_delete(&port_id, &link_id, &tag1.v6_entry.addr).await?;
+
+    tag1.apply_addrs().await?;
+    tag1.verify_addrs_exist(Verify::Exhaustive).await?;
+
+    Ok(())
+}
+
+/// This struct simplifies repetitive CRUD operations
+/// on tagged links with random address registrations.
+struct TestAddrs<'a> {
+    v4_entry: Ipv4Entry,
+    v6_entry: Ipv6Entry,
+    client: &'a Client,
+    port_id: PortId,
+    link_id: LinkId,
+}
+
+impl<'a> TestAddrs<'a> {
+    /// Creates a new instance with a random IPv4 and IPv6 address
+    /// for this port and link.
+    fn new(
+        rng: &mut IpRng,
+        tag: String,
+        client: &'a Client,
+        port_id: PortId,
+        link_id: LinkId,
+    ) -> Self {
+        Self {
+            v4_entry: Ipv4Entry { addr: rng.unique_ipv4(), tag: tag.clone() },
+            v6_entry: Ipv6Entry { addr: rng.unique_ipv6(), tag },
+            client,
+            port_id,
+            link_id,
+        }
+    }
+
+    /// Adds both tagged addresses to this link using dpd's `link_*_create` endpoints.
+    async fn create_addrs(&self) -> anyhow::Result<()> {
+        self.client
+            .link_ipv4_create(&self.port_id, &self.link_id, &self.v4_entry)
+            .await?;
+        self.client
+            .link_ipv6_create(&self.port_id, &self.link_id, &self.v6_entry)
+            .await?;
+        Ok(())
+    }
+
+    /// Adds both tagged addresses to this link using dpd's `port_settings_apply` endpoint.
+    async fn apply_addrs(
+        &self,
+    ) -> Result<(), dpd_client::Error<dpd_client::types::Error>> {
+        self.client
+            .port_settings_apply(
+                &self.port_id,
+                Some(&self.v4_entry.tag),
+                &PortSettings {
+                    links: HashMap::from([(
+                        self.link_id.to_string(),
+                        LinkSettings {
+                            params: LINK_CREATE,
+                            addrs: vec![
+                                self.v4_entry.addr.into(),
+                                self.v6_entry.addr.into(),
+                            ],
+                        },
+                    )]),
+                },
+            )
+            .await?;
+
+        Ok(())
+    }
+
+    /// Fetches this tag's addresses using `link_*_list` and `port_settings_get`.
+    /// Returns Err if both addresses are not found.
+    /// If `scope == Verify::Exhaustive`, returns Err if other addresses
+    /// are found on the link besides those in `self`.
+    async fn verify_addrs_exist(&self, scope: Verify) -> anyhow::Result<()> {
+        let v4 = self
+            .client
+            .link_ipv4_list(&self.port_id, &self.link_id, None, None)
+            .await?
+            .into_inner();
+        let v6 = self
+            .client
+            .link_ipv6_list(&self.port_id, &self.link_id, None, None)
+            .await?
+            .into_inner();
+
+        if !v4.items.contains(&self.v4_entry) {
+            bail!(
+                "Entry {:?} not found in listed addresses: {:?}",
+                self.v4_entry,
+                v4.items
+            );
+        }
+
+        if !v6.items.contains(&self.v6_entry) {
+            bail!(
+                "Entry {:?} not found in listed addresses: {:?}",
+                self.v6_entry,
+                v6.items
+            );
+        }
+
+        if scope == Verify::Exhaustive {
+            anyhow::ensure!(&v4.items == std::slice::from_ref(&self.v4_entry));
+        }
+
+        if scope == Verify::Exhaustive {
+            anyhow::ensure!(&v6.items == std::slice::from_ref(&self.v6_entry));
+        }
+
+        // Verify the port_settings endpoint returns the same.
+        let mut settings = self
+            .client
+            .port_settings_get(&self.port_id, Some(&self.v4_entry.tag))
+            .await?
+            .into_inner();
+
+        let Some(mut settings) =
+            settings.links.remove(&self.link_id.to_string()).map(|s| s.addrs)
+        else {
+            bail!(
+                "port_settings_get should return the target link id({:?}): found {settings:?}",
+                self.link_id
+            );
+        };
+
+        let mut listed = v4
+            .items
+            .into_iter()
+            .filter_map(|entry| {
+                (entry.tag == self.v4_entry.tag)
+                    .then(|| IpAddr::from(entry.addr))
+            })
+            .chain(v6.items.into_iter().filter_map(|entry| {
+                (entry.tag == self.v6_entry.tag)
+                    .then(|| IpAddr::from(entry.addr))
+            }))
+            .collect::<Vec<_>>();
+
+        listed.sort();
+        settings.sort();
+
+        if listed != settings {
+            bail!(
+                "Tagged address sources disagree: link_*_list({listed:?}) v. port_settings_get({settings:?})",
+            );
+        }
+
+        Ok(())
+    }
+}
+
+/// Informs the behavior of address registration verification.
+#[derive(Debug, PartialEq, Eq)]
+enum Verify {
+    /// Expect that the target resources are the only of their
+    /// kind on this link regardless of tag.
+    Exhaustive,
+
+    /// Expect that the target resources exist on the link, but
+    /// resources from other tags may also exist.
+    NonExhaustive,
 }
--- a/dpd-client/tests/chaos_tests/util.rs
+++ b/dpd-client/tests/chaos_tests/util.rs
@@ -4,9 +4,19 @@
 //
 // Copyright 2025 Oxide Computer Company
 
-use dpd_client::Client;
-use dpd_client::types::{Ipv4Entry, Ipv6Entry};
+use std::collections::HashSet;
+use std::net::IpAddr;
+use std::net::Ipv4Addr;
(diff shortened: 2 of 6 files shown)


dpd-client/tests/chaos_tests/port_settings.rs
```diff
@@ -552,3 +516,190 @@ fn random_port_settings() -> PortSettings {
...
+/// Validates address registration namespaces for the following sequence:
+///
+/// - Create link
+/// - Add some addresses to the link manually under tag1
+/// - Port settings apply some different addresses under tag2
+/// - All addresses should be on the link
+/// - Port settings apply away the tag2 addresses
+/// - The tag1 addresses should still be on the link
+#[tokio::test]
+async fn addr_ns_persistent_create() -> anyhow::Result<()> {
```
cfzimmerman, line 525: This failed before the second commit. `addr_ns_spot_delete` passed before changes.

cfzimmerman (commented):
# 🤖 CLAUDE REVIEW

# Review: dendrite#370 — Filter addresses by tag in `LinkSpec` constructor

Base: `main` @ `2475028`. Head: `881403e`. Two commits, 4 files.

The core change is right: making `LinkSpec` tag-scoped stops a `port_settings_apply`
from one client silently deleting another client's addresses, and it also fixes a
latent rollback bug (`addr_del_v4`'s unwind used to re-create a foreign address
stamped with the *applier's* tag, losing the original owner). `cargo check -p dpd
--all-targets` and `cargo check -p dpd-client --features chaos --tests` both pass on
helios.

My concern is that the namespace is only half-built. `apply` and `clear` are now
tag-scoped, but `get`, link deletion, and the per-address endpoints are not — and the
one in-tree consumer that does read-compare-write against this API (omicron's
`DpdPortReconciler`) reads through the un-scoped side.

---

## 1. `get_port_settings` takes a tag and ignores it — omicron's port reconciler can no longer converge

`dpd/src/port_settings.rs:703`

`get_port_settings` now threads a `tag` into the `Context`, but the only consumer of
`ctx.tag` on the read path is `Context::link_spec`, which
`get_port_settings_locked` never calls. It builds its reply from
`LinkSettings::from(&*link)` (`dpd/src/api_server.rs:2952`), which unions **every**
`link.ipv4` / `link.ipv6` entry regardless of tag. So after this PR the read side and
the write side of the same endpoint family disagree about what "this tag's settings"
means.

That asymmetry lands directly on omicron. `PortReconciler::dpd_get_current_settings`
(`sled-agent/scrimlet-reconcilers/src/dpd_reconciler/port_reconciler.rs:131`) calls
`port_settings_get(port, OMICRON_DPD_TAG)`, converts the reply to
`DiffablePortSettings` (dropping only unicast link-locals), and diffs it against
`RackNetworkConfig`. Concretely, with any foreign-tagged non-link-local address on a
managed link:

- **Non-convergence.** `get` reports the foreign address, the diff says "drift",
  the port goes into `to_apply`, and the follow-up
  `port_settings_apply(port, OMICRON_DPD_TAG, …)` now computes its diff over the
  tag-filtered spec, finds nothing to do, and returns `Ok`. Next reconciliation
  pass: identical. The port never appears in `unchanged`, and dpd gets a settings
  apply on every pass, forever. Before this PR the apply deleted the foreign
  address and the loop terminated.
- **Hard failure**, if the foreign address happens to be one omicron *wants*. The
  filtered switch spec omits it, so the diff plans an `add`;
  `create_ipv4_address_locked` (`dpd/src/link.rs:1052`) tests `link.ipv4.contains()`
  and `Ipv4Entry`'s `PartialEq`/`Ord` compare **address only**
  (`dpd-types/versions/src/impls/port.rs:33-52`), so it returns `DpdError::Exists`,
  the whole transaction rolls back, and the port sits in `apply_failures` permanently.

Fix: filter by `ctx.tag` in `get_port_settings_locked`, or drop the parameter and be
explicit that `get` is unscoped. Either way it should not be silently accepted and
discarded — and whichever you pick, the omicron reconciler's expectations want a
second look before this lands in a release omicron picks up.

## 2. Link deletion is still un-scoped, so the tag filter leaks ASIC table entries

`dpd/src/port_settings.rs:378`

`calculate_links` still derives `links_to_del` from `ctx.link_map.port_links(port_id)`
— every link on the port, no tag involved — but `remove_link` now walks a
**tag-filtered** `spec.ipv4`/`spec.ipv6`. So the link goes away while the addresses
belonging to other tags never get `port_ip::ipv4_delete`/`ipv6_delete` called on them.

Nothing downstream picks up the slack. `remove_link` sets `delete_me` and triggers
the reconciler; `link_reconcile` (`dpd/src/link.rs:1813`, `:1871`) calls
`unplumb_link` — which clears the uplink, MAC and ASIC port but never touches
`PortAddrIpv4`/`PortAddrIpv6` — and then `LinkMap::delete_link`
(`dpd/src/link.rs:98`), which is a bare `HashMap::remove`. Note that
`Switch::delete_link` (`dpd/src/link.rs:704`) *does* drain the tables first; the
port_settings path bypasses it.

Reachable path today: tfportd puts a link-local IPv6 tagged `tfportd` on every link
it sees (`tfportd/src/ports.rs:257`). Omicron calls
`port_settings_clear(qsfp, OMICRON_DPD_TAG)` when a port leaves `RackNetworkConfig`
(`port_reconciler.rs:183`), and wicketd's preflight teardown applies an empty
`PortSettings` under the same tag
(`wicketd/src/preflight_check/uplink.rs:808`). Both now delete the link and leave
tfportd's two table entries programmed with no owning `Link`. For v6 the leaked
`drop_key` is `{dst_addr, in_port: val 0, mask 0}` (`dpd/src/table/port_ip.rs:83`) —
a switch-wide blackhole for that address that survives until dpd restarts, and that
will collide if the link is recreated and tfportd re-adds the same address.

Pre-PR the unfiltered spec deleted everything, so this is a regression introduced
here, not a pre-existing hole. Either scope link deletion to the tag too, or have
`remove_link` drain all addresses (not just the tag's) since the link itself is going
away.

## 3. A foreign-tagged address in the desired set now fails the whole apply

`dpd/src/port_settings.rs:113`, `dpd/src/port_settings.rs:518`

Called out inline above but worth its own line, because the fix is different. Because
address identity ignores the tag, `LinkSpec::from_link` filtering it out does not make
the address available — it just makes the diff plan an `add` that is guaranteed to
fail. There's no way for a client to tell "someone else owns this address" from a
generic `Exists`, and the failure takes the entire atomic apply with it. Worth
deciding explicitly what should happen when two tags want the same address: a
distinct error naming the owning tag, or treating a same-address different-tag entry
as satisfied. Right now it's an accident of `Ipv4Entry: Eq`.

## 4. The per-address endpoints are not tag-scoped, so the namespace isn't enforced

`dpd/src/link.rs:1102`, `dpd/src/link.rs:1204`

`delete_ipv4_address_locked`/`delete_ipv6_address_locked` build
`Ipv4Entry { tag: String::new(), addr }` and rely on address-only equality, so
`link_ipv4_delete`/`link_ipv6_delete` will happily remove an address owned by another
tag. tfportd already does exactly this: `ensure_address_match`
(`tfportd/src/ports.rs:240`) deletes whatever `link_local` dpd reports for a link,
with no tag check.

That's pre-existing, but this PR is the one establishing "addresses are namespaced by
tag", and the new `addr_ns_spot_delete` test only ever deletes its *own* addresses —
so it reads as if the property is verified when the enforcement isn't there. Either
scope the delete endpoints or say plainly in the API docs that tags are advisory
outside of `port_settings`.

## 5. The tag's meaning changed for `port_settings_*` but the API docs and versioning didn't

`dpd/src/api_server.rs:1845`, `dpd-api/src/lib.rs:1649-1741`

`query.tag` was previously write-only — a label stamped on entries this call created.
It is now also a *filter* that determines which existing state the call can even see.
`unwrap_or("")` additionally collapses "no tag supplied" into "the empty-tag
namespace", so a caller that omits the parameter now manages only untagged addresses
instead of all of them. That's a material behavioural change to
`port_settings_apply`/`clear`/`get` with no doc-comment update on any of the three
endpoints and no version bump (the schema is unchanged, so the generated clients
won't notice). At minimum the doc comments should state that the tag scopes the
address set.

## 6. Test coverage misses the case that actually breaks (finding 2)

`dpd-client/tests/chaos_tests/port_settings.rs:532`

`addr_ns_persistent_create` deliberately keeps the link alive across the second apply
(it passes the link with `addrs: Vec::new()`), and `addr_ns_spot_delete` never removes
a link at all. Neither test exercises `port_settings_clear`, nor an apply that drops
the link from the settings map — which is precisely where the isolation falls over.
A third case (tag1 addresses on a link, tag2 clears the port, assert tag1's ASIC
entries are gone rather than orphaned) would have caught finding 2. Worth adding
alongside the fix.

## 7. `IpRng` draws from the whole address space

`dpd-client/tests/chaos_tests/util.rs:52`

`Ipv4Addr::from_bits(rng.random())` / `Ipv6Addr::from_bits(rng.random())` can produce
`0.0.0.0`, `127/8`, `224/4`, `::`, `ff00::/8`, or a `fe80::/10` link-local. dpd
doesn't validate any of these today (`link_ipv4_create` goes straight to
`create_ipv4_address`), so the current seeds are fine — but the values are only stable
for this `rand` version, and a `rand` uprev silently redraws every address. Link-local
v6 in particular is special-cased by tfportd and by omicron's diff. Cheap insurance:
generate within `203.0.113.0/24` and `fd00::/8` like the rest of the file does, or
mask the high bits.

## Nits

- `dpd-client/tests/chaos_tests/port_settings.rs:12-13`: the file already imports
  `init_harness` via `use super::harness::{…}` and `link_list_ipv4` via
  `use super::util::{…}`; the two new `use crate::chaos_tests::…` lines plus
  `harness::init_harness(…)` at the call sites are a second, inconsistent style in
  the same file.
- `addr_ns_spot_delete`'s doc comment has a stray sentence break: "This should fail.
  because the address already exists".
- `addr_ns_spot_delete` asserts the duplicate-tag rejection for v4 only; the v6 case
  goes through a separate code path (`create_ipv6_address_locked`) and is untested.

dpd/src/link.rs
```diff
@@ -487,10 +502,7 @@ impl Link {
 
     /// Return the link-local address for this link, if one has been added.
     pub fn link_local(&self) -> Option<Ipv6Addr> {
-        self.ipv6
-            .iter()
-            .find(|entry| (entry.addr.segments()[0] & 0xffc0) == 0xfe80)
-            .map(|entry| entry.addr)
+        self.ipv6.keys().find(|addr| addr.is_unicast_link_local()).copied()
```
cfzimmerman, line 506: This change isn't strictly necessary, but the previous closure did exactly what `is_unicast_link_local` does. I don't see a compelling reason not to use `std` here.

cfzimmerman (commented):
🤖 This message was written by CLU

Review of the tag-isolation fix, **revised after a second pass at `ed381675`**. Everything below describes the PR as it stands now; the threads on this review record how it got there.

The core change — `BTreeSet<Ipv*Entry>` to `BTreeMap<Ipv*Addr, String>` — is the right call: it makes "one address has exactly one owner" a property of the type instead of a set whose element equality quietly ignored half the element. That old shape is what made #342 possible, and the `range()` rewrite of the paginated list falls out much cleaner as a result. Carrying the owner tag on each `LinkSpec` entry, so an apply can reconstruct other tags' addresses instead of overwriting them, fixes the modify path without introducing a second source of truth.

It is worth recording that the headline fix is real and observable in the consumer: `omicron/sled-agent/scrimlet-reconcilers/src/dpd_reconciler/port_reconciler.rs` strips link-local v6 out of its change computation with the comment "dont consider link local addresses in change computation" — a workaround for exactly the pre-PR behavior where a tagged apply clobbered tfportd's `fe80::`. Post-PR that filter should be redundant, since sled-agent already calls `port_settings_get(port, DPD_TAG)`.

## Where this stands

The four findings I wanted resolved before this leaves draft are resolved — two by code, two by a decision that is now written down and pinned by a test.

- **A caller's claim on another tag's address is now an error, not a silent no-op.** `LinkSpec::try_from_settings` errors on the colliding insert, during diff calculation and therefore before any ASIC write, so there is nothing to unwind. `AddrTagConflict` maps to 409, the status the spot-add path already returned for this conflict — though the two still return different errors under it, which is a thread below. `apply_fails_on_tag_conflict` exercises v4 and v6 independently.
- **A cross-tag delete is an error instead of a silent success.** `delete_ipv*_address_locked` returns `AddrTagConflict` naming the owner, symmetrically in v4 and v6, and the rollback hazard behind it is closed: `addr_del_v*` reads the owner tag out of `before` and hands that same tag to the compensating `create`.
- **Link deletion stays tag-blind, deliberately.** A tagged apply that omits a link still destroys it and every other tag's addresses on it. The argument for that is link lifecycle is owned by one client, and the code now says so at `port_settings.rs:229-233` with `apply_implicit_delete` documenting the behavior in a test. `wicketd`'s preflight teardown is a real caller of exactly this path, and it is legitimate under that model.
- **`port_settings_get` stays tag-filtered,** so an untagged GET returns an empty address list. No untagged caller exists: sled-agent's reconciler always passes `DPD_TAG`, wicketd passes `OMICRON_DPD_TAG`, and `swadm` does not expose the endpoint. `link_ipv*_list` remains the unfiltered view.

The fixes themselves are symmetric across v4 and v6, and the conflict check precedes mutation rather than relying on rollback — I checked both, because a fix that lands on one address family or unwinds badly is the usual way a second version goes wrong.

## Still outstanding

**`dpd/src/link.rs:55` fails CI.** `use std::fmt::Debug;` is new and unused, and `cargo clippy -- --deny warnings` runs in all four feature legs. One line; it has its own thread.

**The API docs say nothing about tags, and by now they need to.** `port_settings_apply`, `port_settings_clear` and `port_settings_get` carry one line of prose each, none of which mentions the word. A caller integrating against this cannot discover any of the rules this PR just established:

- an apply is scoped to the caller's tag for *addresses*, but owns the port's link set outright — omitting a link deletes it and everything on it;
- an apply that names an address another tag owns fails with 409 rather than taking it;
- `tag` is typed `Option<String>`, and absent means the tag `""` — which is a distinct owner, not "all owners", so an untagged GET returns nothing.

Three sentences on the trait, and the last real gap in this change closes.

Smaller things live in the threads rather than here: `apply_fails_on_tag_conflict` never asserts that the other tag's addresses survived the apply it rejected, which is the property #342 is about; the negative assertions elsewhere pass for reasons other than the deletion they claim to check; and the spot-add path still reports a cross-tag collision as `Exists` where the bulk path now reports `AddrTagConflict`.

**1. What customer problem is this actually solving?** A concrete one, and the workaround in omicron is the proof. `tfportd` owns link-local addresses and `sled-agent` owns routable ones on the same links; before this, a tagged `port_settings_apply` from either deleted the other's. On a real rack that is a link-local address disappearing out from under ddm mid-reconcile, with no error anywhere — the daemon that lost its address quietly stops working until its next pass. sled-agent's response was to stop looking at link-local addresses entirely, which is a reasonable patch on a bug it could not fix from its side. Fixing it properly in dpd lets that filter be deleted.

**2. Does the value warrant the complexity?** More comfortably now than at the first pass. The map change removes a concept rather than adding one. The cost I flagged sat in `from_settings`, whose two non-obvious rules — *the result contains addresses the caller never mentioned*, and *a claim on a foreign address does not win* — were encoded in an `extend` with a filter and stated nowhere. Both are now explicit: the function is `try_from_settings`, `master` is `current`, the doc says what is preserved and what errors, and the second rule is a typed error rather than a silent precedence. That was the whole of my complexity objection and it is gone.

**3. What is the simplest robust alternative?** Unchanged, and still this one. *Do nothing* is not viable — the bug is filed and already worked around downstream. *Fix the caller* does not exist as an option: the entire premise of the tag is that the callers are mutually unaware daemons, so dpd is the only place this can be fixed. *Make the tag a real type* remains the right next step and is still correctly deferred — and the second pass sharpened why. A `Tag` newtype forces a decision about the `""` sentinel: is untagged a distinct owner, or does it mean "anyone"? This PR has now answered that in practice — a distinct owner named empty-string, which is why an untagged GET returns nothing and why a tagged client can never adopt an untagged address. That answer is defensible, but it is currently expressed only as `unwrap_or("")` in three handlers. Naming it is the prerequisite for the newtype.

**The uncomfortable question, answered.** At the first pass, four findings shared one pattern: the server did less than the caller asked and reported success. Tag isolation only pays off if a client can tell when its request was refused — otherwise you replace "my address got deleted" with "my address was never created", which is quieter but no easier to debug at 3am, and in the reconciler case turns a visible failure into an invisible infinite loop. That pattern is now gone from the address paths: both the bulk and spot paths return 409 with the owner named. It survives in exactly one place, by choice — link deletion — and the remaining ask is only that the API docs say so out loud, since a caller cannot read the comment in `port_settings.rs`.

*Revised 2026-09-11 at `ed3816755dea`.*

dpd/src/port_settings.rs
```diff
@@ -211,21 +239,18 @@ impl PortSettingsDiff {
...
             .map(|id| {
-                let settings_link = (&settings.links[&id.0]).into();
-                let switch_link = ctx.link_spec(*id).expect(
+                let before = ctx.link_spec(*id).expect(
                     "link existence is guaranteed by the locked link map",
                 );
-                (id, settings_link, switch_link)
+                let conf = &settings.links[&id.0];
+                let after =
+                    LinkSpec::from_settings(conf, ctx.tag, Some(&before));
```
cfzimmerman: 🤖 This message was written by Claude

**Link deletion is still tag-blind, so the #342 bug survives one level up.**

This is the fix for the modify path — `from_settings(conf, ctx.tag, Some(&before))` carefully preserves other tags' addresses. But the delete path just above (`links_to_del` at line 218, `self.links.delete` at 229-237) got no equivalent treatment: it is still `switch_links.difference(&settings_links)`, and `remove_link` then iterates the *whole* `spec.ipv4`/`spec.ipv6` map — built from `LinkSpec::from(&Link)`, i.e. every tag's addresses — and deletes them all.

Concretely: `tfportd` creates link 0 on qsfp0 and registers its link-local. `sled-agent` calls `port_settings_apply?tag=sled-agent` with settings naming only link 1. Link 0 is on the switch and absent from sled-agent's body, so it lands in `links.delete` and tfportd's address goes with it. That is #342 through a different door, and the new tests don't cover it — `addr_ns_persistent_create` applies an empty `addrs` for a link that is still *present* in the body, which exercises the modify path only.

There may be no clean answer today, since links carry no tag of their own — only their addresses do. But the options deserve an explicit decision rather than silence: skip deletion of a link whose addresses are all foreign; refuse the apply with a conflict; or state in the API docs that a tagged apply owns the port's link set outright and only addresses are tag-scoped. Note you already updated `port_settings_clear`'s doc to be honest about this ("Deletes all the links belonging to a port") — `port_settings_apply` deserves the same. If it's out of scope, fine, but then #342 shouldn't be closed here without a follow-up issue naming this case.

dpd/src/port_settings.rs
```diff
@@ -114,38 +114,64 @@ impl From<&Link> for LinkSpec {
...
+            );
+
+            ipv6.extend(
+                master
+                    .ipv6
+                    .iter()
+                    .filter(|(_, entry_tag)| entry_tag.as_str() != tag)
+                    .map(|(addr, entry_tag)| (*addr, entry_tag.to_string())),
+            );
+        }
```
cfzimmerman: 🤖 This message was written by Claude

**A caller's claim on an address owned by another tag is silently discarded, which livelocks a reconciler.**

`extend` overwrites on key collision, so ordering matters: every address from `l.addrs` goes in with the caller's `tag` first, then this block overwrites any that a *different* tag already owns. The result is `after[addr] == before[addr]`, the `modify_link` filter (`!ipv4_before.contains_key(addr)`) skips it, and `port_settings_apply` returns `Ok` — while the address the caller asked for is still owned by someone else. The response body, built by `Link::settings(tag)`, omits it entirely.

The failure mode is worse than a lost write. sled-agent's `port_reconciler` diffs desired against what `port_settings_get(port, DPD_TAG)` returns. If the desired config names an address tfportd owns, every pass sees current ≠ desired, re-applies, gets the same silent no-op, and never converges — `ports_to_apply` never drains and nothing is logged. A hard error would surface in one pass; this hides forever.

It's also inconsistent with the spot-add path: `create_ipv4_address_locked` returns `DpdError::Exists` for exactly this conflict, and your own new test asserts that it does. Same conflict, two answers depending on the endpoint. Given the PR summary says "the same address cannot be registered by multiple peers," the bulk path should say so too — collect collisions and fail the apply naming the address and its current owner.

Separately: `master` reads as "the authoritative one," but it holds current switch state. `current` describes it better, and `from_settings` deserves a doc line saying foreign entries win — that rule is not guessable from the signature.

dpd/src/api_server.rs
```diff
@@ -1876,9 +1875,10 @@ impl DpdApi for DpdApiImpl {
         let path = path.into_inner();
         let query = query.into_inner();
         let port_id = path.port_id;
+        let tag = query.tag.as_deref().unwrap_or("");
```
cfzimmerman, line 1878: 🤖 This message was written by Claude

**This collapses "no tag supplied" into "the empty tag", which turns an untagged `port_settings_get` into a blank response.**

Previously `get_port_settings` took the `Option` through and the old `impl From<&Link> for LinkSettings` added every address unconditionally, so a bare `GET /port/qsfp0/settings` showed the whole picture. Now the `None` becomes `""`, and `Link::settings` (line 2958) filters `*t == tag` — so an untagged GET returns only addresses registered with the literal empty tag, which in practice is none of them. An operator or support engineer curling this endpoint to debug a configured port sees an empty `addrs` list on every link, and there is no longer any way to ask for the unfiltered view.

sled-agent is unaffected — it always passes `DPD_TAG` (`port_reconciler.rs:131`) — so this lands on humans and `swadm`, not the control plane. But note the versioning asymmetry: `port_settings_clear` got a v14 bump for its semantic change while this one rides on the existing `VERSION_ALLOW_DDM_TRAFFIC..` range, so a v13 client gets new `port_settings_get` semantics with no signal that anything changed.

Carrying `Option<&str>` through to `Link::settings` and treating `None` as unfiltered keeps the debug view and costs almost nothing. Same `unwrap_or("")` at line 1845 for `port_settings_apply` is fine by comparison — there "" is a real owner to write under, not a filter.

dpd/src/link.rs
```diff
@@ -1082,40 +1073,50 @@ impl Switch {
...
-                address: address.into(),
-            })
+                address: addr.into(),
+            }),
+            btree_map::Entry::Occupied(slot)
+                if tag.is_some_and(|t| t != slot.get()) =>
+            {
+                // Don't delete an addr that belongs to another tag.
+                Ok(())
+            }
```
cfzimmerman, line 1118: 🤖 This message was written by Claude

**Returning `Ok(())` here reports success for a delete that deliberately did nothing — and sets a rollback trap.**

The doc comment above says "Returns Err if the address is not found," but there are two ways to not delete and only one is an error. A caller passing `Some(tag)` cannot distinguish "deleted" from "refused, not yours."

Today every `Some(tag)` caller is internal to `port_settings.rs` and reads the tag out of `before`, so the arm is unreachable — which is exactly what makes it a trap. If it ever does fire inside `addr_del_v4`, that function has already called `rb.wind(... create_ipv4_address_locked(addr, tag))`. A later failure in the same transaction then unwinds into a create against an address that was never removed, which returns `DpdError::Exists` and converts an ordinary recoverable failure into `DpdError::Unwind` / `ROLLBACK_FAILURE_ERROR_CODE`. A silent no-op becomes a poisoned rollback.

Make the outcome explicit: return `DpdError::NoSuchAddress`, add a distinct "owned by another tag" error, or change the signature to `DpdResult<bool>` so callers must look. Same applies to `delete_ipv6_address_locked` at 1223.

dpd-client/tests/chaos_tests/port_settings.rs
```diff
@@ -552,3 +515,325 @@ fn random_port_settings() -> PortSettings {
...
+            .items
+            .into_iter()
+            .filter_map(|entry| {
+                (entry.tag == self.v4_entry.tag)
+                    .then(|| IpAddr::from(entry.addr))
+            })
+            .chain(v6.items.into_iter().filter_map(|entry| {
+                (entry.tag == self.v4_entry.tag)
+                    .then(|| IpAddr::from(entry.addr))
+            }))
```
cfzimmerman, line 805: 🤖 This message was written by Claude

**The v6 branch filters on the v4 entry's tag.**

Line 807 and line 811 both read `self.v4_entry.tag`; the second is inside the IPv6 arm and should be `self.v6_entry.tag`. `TestAddrs::new` gives both entries the same tag, so this passes today and will keep passing — which is the problem. The moment a test wants to check that v4 and v6 registrations under *different* tags stay isolated, this helper will silently report success without testing it, and that is precisely the scenario this PR exists to protect.

While here: `Verify::Exhaustive` asserts `items.len() != 1` rather than comparing against the expected set. It will start failing for a confusing reason as soon as anything else lands on the link — a harness-created link-local, say — and the message won't say what was actually there. Comparing sorted vectors is about the same amount of code and gives a usable failure.

dpd/src/api_server.rs
```diff
@@ -1178,7 +1178,7 @@ impl DpdApi for DpdApiImpl {
         let link_id = path.link_id;
         let address = path.address;
         switch
-            .delete_ipv4_address(port_id, link_id, address)
+            .delete_ipv4_address(port_id, link_id, address, None)
```
cfzimmerman, line 1181: 🤖 This message was written by Claude

**Spot delete passes `None`, so it still bypasses tags entirely.**

This endpoint now has tag plumbing available and explicitly opts out, which leaves `link_ipv4_delete` as a way for any client to delete any other client's address — the same isolation hole as #342, on a different endpoint. The PR body lists spot deletion as fine, and the create side genuinely is (`create_ipv4_address` threads `entry.tag` correctly), but delete takes no tag in the API at all.

I'm not arguing for a behavior change here — adding tag scoping would be an API break needing its own version bump, and `swadm` operators plausibly want the override. But the `None` is doing load-bearing work with no explanation. A short comment saying this endpoint is intentionally tag-agnostic, and why, stops the next reader from filing it as the bug it resembles. Same for the v6 case at line 1256.

dpd-client/tests/chaos_tests/util.rs
```diff
@@ -37,3 +47,64 @@ pub(crate) async fn link_list_ipv6(
...
+        })
+    }
+
+    /// Returns a random IPv6 address that this instance
+    /// has never created before.
+    pub fn unique_ipv6(&mut self) -> Ipv6Addr {
+        Self::roll_unique(&mut self.claimed, || {
+            Ipv6Addr::from_bits(self.rng.random())
+        })
+    }
```
cfzimmerman, line 77: 🤖 This message was written by Claude

**Unconstrained random addresses will eventually generate ones that aren't valid link addresses.**

`Ipv4Addr::from_bits(rng.random())` draws uniformly from the whole 32-bit space, so roughly **1 in 8** generated v4 addresses lands in 224.0.0.0/3 (multicast and reserved), plus smaller odds of 0.0.0.0/8, 127.0.0.0/8 and 255.255.255.255. On the v6 side about 1 in 256 lands in ff00::/8 and about 1 in 1024 in fe80::/10.

The link-local case is the one I'd worry about: it isn't merely invalid, it's *meaningful*. `Link::link_local()` returns the first link-local address on the link, so a randomly drawn `fe80::` can silently become the link's link-local and change behavior unrelated to the test.

It's seeded, so today's runs are deterministic and evidently pass — which makes the seed load-bearing in a way nothing records. Adding a test, reordering existing ones, or changing the seed can start drawing a multicast address and produce a failure that looks nothing like its cause. Drawing from the documentation ranges instead — 198.51.100.0/24 and 2001:db8::/32, randomising host bits — keeps the uniqueness property and removes the class entirely.

(Minor: `#[cfg(test)]` on `mod util_tests` in a `tests/` file is redundant — integration tests are already compiled with `--test`.)

dpd/src/link.rs
```diff
@@ -1047,17 +1036,18 @@ impl Switch {
...
-                entry.addr
-            )))
-        } else {
-            port_ip::ipv4_add(self, link.asic_port_id, entry.addr)?;
-            link.ipv4.insert(entry);
-            Ok(())
+        match link.ipv4.entry(addr) {
+            btree_map::Entry::Occupied(_) => Err(DpdError::Exists(format!(
+                "IP address {addr} already exists",
+            ))),
```
cfzimmerman: 🤖 This message was written by Claude

Nit, but cheap given what this PR is for: now that every address has an owner, `"IP address {addr} already exists"` omits the one fact the operator needs. When sled-agent gets this for an address tfportd owns, the message should say so — `slot.get()` is right there. Same for the v6 copy at 1159.

cfzimmerman:
fyi human reviewers, if not self-evident, I'm experimenting with claude review infrastructure.

dpd/src/port_settings.rs
cfzimmerman: I'll add a comment, but this is not easy to fix. Right now tags only apply to addresses on the link, not the link itself. Pushing ownership deeper into the link adds complexity we don't have a justification for. Address tagging only exists to synchronize tfportd and sled-agent, and implicit *link* deletion is okay because only sled-agent controls that part. We just don't want implicit *address* deletion.

dpd/src/port_settings.rs
cfzimmerman: My understanding of the issue: Single address creation returns `DpdError::Exists` if two tags try to claim the same address. The new port_settings code silently chooses a winner. Instead we need to return error on a tag conflict.

dpd/src/api_server.rs
cfzimmerman: This seems like a reasonable concern. Conclusion: `port_settings_get` with `tag == None` needs to return all addresses.

dpd/src/link.rs
cfzimmerman: This is a valid concern. Solution is to return error when a user tries deleting an address belonging to another tag.

dpd-client/tests/chaos_tests/port_settings.rs
cfzimmerman: Copy paste bug fixed. Addressed `Verify::Exhaustive` as well.

dpd/src/api_server.rs
cfzimmerman: The previous review comments made me realize I need more tests. Those new tests will exercise this tag, so I'll leave this test alone. The other tests will make this tag more self explanatory.

dpd-client/tests/chaos_tests/util.rs
cfzimmerman: I think this is fine. We're not using anything here that depends on the nature of the address, so arbitrary bytes   are conceptually accurate. rng seeding prevents unexpected issues in future test runs.

dpd/src/link.rs
cfzimmerman: Added tag to the error message in v4 and v6.

dpd/src/api_server.rs
cfzimmerman: Actually I think that's not good. Merging tagged addresses in `PortSettings` is exactly what we're trying to escape. The user can still get all link addresses using the `link_ipv*_list` endpoints. The `port_settings_*` endpoints are necessarily tagged. It's more faithful to make the tag non-optional in the API than to return all addresses when `tag == None`.

cfzimmerman: Yuck. I see two options:
1. Update the `port_settings_*` routes to make tag required.
2. Update swadm to make tag required on `port_settings_get`.

(1) is a more complete solution. But this PR has landed zero breaking changes to sled-agent. I'd really like to keep it that way. If what exists today will work as we need, I'd rather not mess with the API right now.

(2) Addresses your example exactly. That's awkward, but afaik there are no edge cases and no slippery slope arguments. Yes it would be better if `None` did not unwrap to `""`, but no it's also not hurting anything in practice as long as swadm enforces the same contract as sled-agent and tfportd.

dpd/src/link.rs
```diff
@@ -59,6 +61,19 @@ use std::time::Duration;
...
+// - IP addresses live in a BTree map/set (not Hash map/set) because the
+//   dropshot API for listing addresses is paginated. BTree gives a stable
+//   way to iterate segments of addresses.
+// - IP address CRUD functions are duplicated and not generic. A generic
+//   v4/v6 design either motivates even more wrapper functions, or it leaks
+//   complexity into the `port_settings` winding and `port_ip` table
+//   handlers. This quickly degrades the LOC reductions here from using
+//   generics. I still think a generic approach is attractive, but the
+//   existing design expresses reasonable skepticism of that marginal benefit.
+
```
cfzimmerman: I'm conflicted on whether a comment like this is warranted. But these were both decisions I questioned until exploring changes and finding a dead end. This is now obvious to me, but maybe it's useful for others to know as well? Open to opinions.

dpd/src/link.rs
```diff
@@ -236,10 +251,11 @@ pub struct Link {
...
     pub media: PortMedia,
-    /// A list of IPv4 addresses assigned to this link.
-    pub ipv4: BTreeSet<Ipv4Entry>,
-    /// A list of IPv6 addresses assigned to this link.
-    pub ipv6: BTreeSet<Ipv6Entry>,
+    /// IPv4 addresses assigned to this link.
+    /// Each is associated with its creator's ID tag.
+    pub ipv4: BTreeMap<Ipv4Addr, String>,
+    /// IPv6 addresses assigned to this link.
+    pub ipv6: BTreeMap<Ipv6Addr, String>,
```
cfzimmerman, line 244: I recently preached to @taspelund about the glorious merits of `Box<str>` in a different PR review, so using `String` here feels a bit hypocritical. Lol sorry. But after considering it more, using `Box<str>` is not cool unless we use it everywhere, which is unlikely because dropshot APIs like `String`. In a mixed case, `Box<str>` just forces realloc upon perfectly fine `String`s.
If I'm going to shed ink over a dendrite `Tag` type, it's not gonna be a `Box<str>` anyway. So might as well stick with convention.

dpd/src/link.rs
```diff
@@ -1047,17 +1036,19 @@ impl Switch {
...
-            Err(DpdError::Exists(format!(
-                "IP address {} already exists",
-                entry.addr
-            )))
-        } else {
-            port_ip::ipv4_add(self, link.asic_port_id, entry.addr)?;
-            link.ipv4.insert(entry);
-            Ok(())
+        match link.ipv4.entry(addr) {
+            btree_map::Entry::Occupied(curr) => Err(DpdError::Exists(format!(
```
cfzimmerman: TODO::cory: use addr conflict error type here

dpd/src/port_settings.rs
```diff
@@ -114,40 +114,79 @@ impl From<&Link> for LinkSpec {
...
 }
 
-impl From<&LinkSettings> for LinkSpec {
-    fn from(l: &LinkSettings) -> Self {
-        Self {
+impl LinkSpec {
+    /// Constructs a [`LinkSpec`] based on the given [`LinkSettings`].
+    ///
+    /// If `current` is Some, the resulting LinkSpec includes all
+    /// tagged resources belonging to *other* tags.
```
cfzimmerman: TODO::cory: confusing comment

dpd/src/link.rs
```diff
@@ -1240,7 +1261,7 @@ impl Switch {
         link_id: LinkId,
     ) -> DpdResult<()> {
         self.link_update(port_id, link_id, |link| {
-            while let Some(Ipv6Entry { addr, .. }) = link.ipv6.pop_first() {
+            while let Some((addr, _)) = link.ipv6.pop_first() {
                 port_ip::ipv6_delete(self, link.asic_port_id, addr)?;
             }
```
cfzimmerman: TODO::cory: memory leak?

dpd/src/api_server.rs
cfzimmerman: Looks like `swadm` doesn't even expose `port_settings_get`. I think `sled-agent` is the only real consumer, and it uses a tag.

https://github.com/search?q=org%3Aoxidecomputer+port_settings_get&type=code&p=1

dpd/src/link.rs
```diff
@@ -50,7 +50,9 @@ use slog::o;
 use slog::warn;
 use std::collections::BTreeMap;
 use std::collections::BTreeSet;
+use std::collections::btree_map;
 use std::collections::btree_map::Entry;
+use std::fmt::Debug;
```
cfzimmerman: 🤖 This message was written by CLU

**Category: Demonstrable bug**

**This import is unused, and CI denies warnings.**

The only other `Debug` occurrences in this file are four `#[derive(Debug)]` at lines 227, 323, 366 and 419, and a derive expands to an absolute path — it does not need the trait in scope. Nothing here takes a `Debug` bound.

`.github/buildomat/jobs/test.sh` runs `cargo clippy --features $feat -- --deny warnings` for `tofino_stub`, `tofino_asic`, `softnpu` and `chaos`, so `unused_imports` fails the build in all four legs before the test step is reached.

I have not compiled the branch — this is read off the file and the CI script, both at `ed381675`.

cfzimmerman (commented):
🤖 This message was written by CLU

Follow-up pass over the second version (`ed381675`). One new finding, below — everything raised in the first pass is answered in the threads on [the earlier review](https://github.com/oxidecomputer/dendrite/pull/370#pullrequestreview-5149037559), whose summary I have revised to describe the PR as it stands now rather than as it was.

**Findings:** 1 demonstrable bug

dpd/src/port_settings.rs
cfzimmerman: 🤖 This message was written by CLU

**Addressed — safe to resolve.**

Agreed, and the shape you landed is better than what I asked for. The decision is recorded at `port_settings.rs:229-233`, and `apply_implicit_delete` pins the behavior in a test that says out loud it is documenting a footgun rather than endorsing it — that is the explicit decision the thread was asking for, plus a tripwire if it ever changes silently.

I checked the one caller that actually exercises this: `wicketd`'s preflight teardown (`omicron/wicketd/src/preflight_check/uplink.rs:808`) applies an empty `PortSettings` under `OMICRON_DPD_TAG` specifically to delete the link it created. That is legitimate under your model — omicron owns the link lifecycle, so taking its addresses with it is correct, not collateral.

The one piece still missing is a sentence in the *API* docs rather than in the source, since a caller cannot read `port_settings.rs`. That spans three endpoints so I have moved it to the review summary instead of leaving this thread open.

One thing worth naming while this closes, because it is less obvious than deletion: the same hole covers link *parameters*. `try_from_settings` takes `speed`, `fec`, `autoneg`, `kr`, `tx_eq` and `allow_ddm_traffic` from the caller's `LinkSettings` unconditionally, so a tagged apply from tfportd would preserve sled-agent's addresses and still silently reconfigure sled-agent's link. Your comment's first sentence — "tags only protect address modifications" — already says this; the example that follows it is only about deletion, and the parameter case is the one a reader is less likely to derive for themselves.

dpd/src/port_settings.rs
cfzimmerman: 🤖 This message was written by CLU

**Addressed — safe to resolve.**

Fixed, and fixed at the right layer. `LinkSpec::try_from_settings` seeds the map with other tags' entries and then errors on a colliding insert, so the conflict is detected during `PortSettingsDiff::calculate` — before any ASIC write — and there is nothing to unwind. `AddrTagConflict` maps to 409, which matches what `create_ipv4_address_locked` already returned for the same conflict, so the bulk and spot paths finally agree.

`apply_fails_on_tag_conflict` covers v4 and v6 independently, which is the part I would have worried about: it deletes the v4 address and re-applies to prove the v6 arm errors on its own rather than riding on the v4 check.

The naming and doc points landed too — `master` is now `current`, and the doc comment states that foreign entries are preserved and that a claim on another tag's address is an error.

Two notes on the test rather than the fix. Its doc says "successful rollback is important in such cases", but no rollback runs in it — the conflict is raised in `calculate`, before `execute` — so the comment points a future reader at the wrong property. And after each `expect_err` it never checks that TAG2's addresses are still there, which is the property #342 is actually about; a `tag2.verify_addrs_exist(...)` at each of those points would make it a test of isolation rather than of the status code.

Worth stating the trade this makes explicit: an apply that used to no-op now fails permanently. That is the right direction — a reconciler that cannot converge should say so — and sled-agent should not hit it, since its desired set comes from uplink config and its change computation already skips link-local addresses. But it is a behavior change only omicron can really validate.

dpd/src/api_server.rs
cfzimmerman: 🤖 This message was written by CLU

**Addressed — safe to resolve.**

Your reasoning holds and I checked the premise rather than taking it: across omicron, the only `port_settings_get` caller is sled-agent's `port_reconciler.rs:131`, which always passes `DPD_TAG`, and `wicketd`'s preflight uses `port_settings_apply` with `Some(OMICRON_DPD_TAG)`. Nothing calls it untagged, and `link_ipv*_list` remains the unfiltered view, so "the `port_settings_*` endpoints are inherently tagged" is a position the code actually supports.

One correction to my original comment: the versioning asymmetry I cited is not there at this head. `port_settings_apply`, `_clear` and `_get` all ride `VERSION_ALLOW_DDM_TRAFFIC..` with a single `_v1` shim below it, and there is no `port_settings_clear_v2`. So there is no bump to be inconsistent with.

What is left is only that the API still types `tag` as `Option<String>` while the server now reads absent as the tag `""` — a documentation gap, not a behavior one. I have folded that into the summary with the other docs point.

dpd/src/link.rs
cfzimmerman: 🤖 This message was written by CLU

**Addressed — safe to resolve.**

Fixed, symmetrically in v4 and v6, and the error names the owner so the caller can act on it. The doc now states both outcomes — `NoSuchAddress` when absent, `AddrTagConflict` when it belongs to someone else, delete-regardless when `tag` is `None`.

The rollback trap is closed as a side effect of the `port_settings` change: `addr_del_v4` now reads the owner tag out of `before` and passes that same tag to both the delete and the compensating `create` in `rb.wind`, so the wind/unwind pair can no longer disagree about ownership.

dpd-client/tests/chaos_tests/port_settings.rs
cfzimmerman: 🤖 This message was written by CLU

**Addressed — safe to resolve.**

Both fixed — the v6 arm filters on `v6_entry.tag` now, and `Verify::Exhaustive` compares the listed items against `std::slice::from_ref(&self.v*_entry)` instead of asserting on a length.

Two leftovers, neither worth holding this up. The `ensure!` calls carry no message, so an exhaustive failure prints the expression rather than what was actually on the link — the two `bail!`s above them do it right and cost the same. And `TestAddrs` still reaches through `v4_entry.tag` for the struct's tag in `apply_addrs` and in the `port_settings_get` call; that is safe while `new()` hands both entries the same tag, but a single `tag` field would remove the class of bug this thread was about rather than relocating it.

Two more in the same helper, same weight. The negative assertions spell "these addresses are gone" as `verify_addrs_exist(...).expect_err(...)`, which also passes on a 404 from `link_ipv*_list`, a link missing from `port_settings_get`, or any transport error — in `apply_implicit_delete` the link may be either gone or present-and-empty and both paths pass. And the two `if scope == Verify::Exhaustive` blocks are back to back and can be one.

dpd/src/api_server.rs
cfzimmerman: 🤖 This message was written by CLU

**Addressed — safe to resolve.**

Good enough, and the explanation ended up in a better place than the call site: `delete_ipv*_address_locked` now documents that `None` deletes regardless of tag, which is what a reader following the `None` will actually open. `deleted_address_must_exist` says the same thing in the test.

One note for the record rather than a request: that test's comment reads "Deletion is not currently tagged, so there's no way/reason to test tag conflicts", and the untested path is precisely the one the `None` enables — one client deleting another client's address through `link_ipv*_delete`. Deliberate per the PR, just worth knowing it is deliberate-and-uncovered rather than covered.

Same thought applies to the doc on `address_cmds_respect_tags`: it says create *and* delete don't affect other tags' resources, but delete is exactly the endpoint that ignores tags, and the test only ever deletes tag1's own addresses. Narrowing that sentence to creation would keep the test honest about what it covers.

dpd/src/link.rs
cfzimmerman: 🤖 This message was written by CLU

**Addressed — safe to resolve.**

Done in both — `IP address {addr} already exists under tag {}` reads off the occupied entry, so the message now names the daemon holding it.

Two follow-ons that only became visible once the bulk path grew its own error. `create_ipv*_address_locked` still answers a cross-tag collision with `Exists`, while `port_settings_apply` now answers the same event with `AddrTagConflict` — both 409, so a client on the spot-add path cannot tell "I already own this" (benign, safe to retry) from "another tag owns this" (needs a human) without parsing the message. Returning `AddrTagConflict` when `curr.get() != tag` and keeping `Exists` for a same-tag re-create would close that.

Also worth knowing where the improved text lands: `DpdError::Exists` maps through `HttpError::for_client_error_with_status(Some(e), CONFLICT)`, whose first argument is the **`error_code`** — dropshot sets the external message to the status reason, "Conflict". So the tag now appears in `error_code`, a field a client may be matching on, rather than in the message. `AddrTagConflict` sets `external_message` directly, which is the better of the two shapes.

dpd/src/port_settings.rs
cfzimmerman: Changing API docs has arisen here and elsewhere. My goal with this PR is to not require a new dropshot version. Updating the API would endorse or formalize what we have. I'm just trying to make legacy code correct. I'd rather save API changes for a design discussion about what we really want. When we have that conversation, the new unit tests elucidate what's happening about as well as docs.

dpd/src/link.rs
cfzimmerman: Nice. There's no rustc/clippy warning, I think because it's implicitly used. But prelude seems to do the job just as well. Removed.

cfzimmerman:
Btw once this merges, `sled-agent` should not need this anymore: https://github.com/oxidecomputer/omicron/blob/6652b9a3a321cb306ea0f2e2328ac789f829f098/sled-agent/scrimlet-reconcilers/src/dpd_reconciler/port_reconciler.rs#L466-L474

It's only required now because sled agent is seeing addresses tagged to tfportd.

@andrewjstone fyi, as a friend and ally of `port_settings_*` functions.

dpd-client/tests/chaos_tests/port_settings.rs
```diff
@@ -552,3 +518,527 @@ fn random_port_settings() -> PortSettings {
...
+    Ok(())
+}
+
+/// Verifies the current known behavior of port_settings_clear, which
+/// is to delete the link and all its config from the port.
+///
+/// This test isn't necessarily endorsing the implementation, but
+/// it does seek to track API's current behavior.
+#[tokio::test]
+async fn settings_clear_ignores_tags() -> anyhow::Result<()> {
```
cfzimmerman, line 640: This is a strange behavior. But I also don't see a clearly better design under the current API.
- Neither the link nor most of its properties are tagged. So `port_settings_clear` could delete only the addresses belonging to the tag. But what should it do about non-tagged settings? The current solution of just dropping the link conveniently avoids deciding or caching what a link's default settings should be.
- We could reject "clear" if a link has multiple tags. But that makes this route useless because we expect links to often have multiple tags.
- We could delete this and have caller use `link_delete` instead. I like this option, but I didn't want this PR to have breaking API changes. The goal right now is just fixing internal issues.

PR #268, opened 2026-05-10
bump v6 lpm size

rcgoodfellow:
Brings the V6 routing tables up to 8291 entries, in line with V4 routing tables.

Something to note is that `Ingress.l3_router.Router6.lookup_idx.lookup` is now [split across 3 stages](https://buildomat.eng.oxide.computer/wg/0/artefact/01KR87Z12ZKA1BKSFR1P8QK635/0q7MSkcSxtZU1fLueTUEN7zmyAC8qlPy9maNOcYjenuVo7qK/01KR87Z99D3005VR1ZY9BFK48V/01KR88PTBPCFNP814YT6KK1CRA/mau.characterize.log).

PR #371, opened 2026-09-09
[mcast] remove multicast feature + keep routing table sizes

zeeshanlakhani:
This work is split up into two pieces in turning multicast "on" by default: 

## [mcast, feature] Feature removal

This removes the multicast feature gating altogether, making multicast inclusive to the work in dendrite/dpd/sidecar.

## [p4, sidecar] keep (IPV*_LPM_SIZE - 1) entries for routing tables w/ multicast

Removing the multicast feature gate got us over Tofino's 20-stage pipeline limit. The table resizing work in #268 matched the v6 LPM table size to what we had in v4, putting us at 20 stages without multicast.

In #208, I explored a few different approaches torefactor our pipeline, across ingress and egress, to give us more breathing room (stage-wise). However, we want to be efficient and targeted when making changes to the sidecar.

Reducing mcast table sizes was my first attempt. Route and replication tables fit in SRAM; SSM is TCAM. The result, sadly, was fewer TCAM blocks, but the same stage count.

Instead, I went the direction of fulfilling a Nils'ism:

> XXX: this control could be moved to the Egress pipeline if we need more space
> in the Ingress pipeline.  Currently unicast packets are able to bypass that
> pipeline, which is why we've tacked it on here.  We could probably also merge
> it with the MacRewrite control, as they are both per-port settings, but that
> would present some weird semantics to the control plane daemon.

I moved the `EgressFilter` control to the egress pipeline to save a stage, carrying (or bridging) the NAT egress bit in the bridge header. I *did not* move the `MacRewrite` control over to handle both uni+multi(cast) in egress, but that was demonstrated in #208 for a follow-up (post-testing).

With the move, 20 stages is achieved! However, the v4/v6 route tests still stopped below 8191 entries with a deficit. The exact-match route-target tables were losing capacity under four-way cuckoo placement (default). [`@ways(8)`][ways] was the trick I found that gives each key in the table eight placement choices, which now makes table tests fill all 8191 entries.

Includes:
  - Removal of the `#ifdef MULTICAST` codegen guards.
  - Counter updates to match our move of the NAT egress filter

[ways]: https://github.com/p4lang/p4c/blob/a19f1c3d85a867a6288fd983f7bad505ac47d728/backends/tofino/bf-p4c/common/pragma/pragmas.cpp#L1172-L1183


dpd-client/tests/integration_tests/table_tests.rs
```diff
@@ -33,26 +32,17 @@ use crate::integration_tests::common::prelude::*;
...
+// LPM lookups live in TCAM, placing entries without hash conflicts.
+// The exact-match index->target tables behind them are cuckoo-managed SRAM
+// (see pipe_mgr/cuckoo_move.h in p4lang/open-p4studio). Without enough
+// hash ways, cuckoo conflicts can cap capacity a few % below what's
+// declared. The @ways(8) annotation on those route tables helps us reach full
+// occupancy on this build. It is normal for these sizes to change slightly
+// after updating the P4 code. If the table size appears to change
+// dramatically, that's worth investigating. If it only changes by an
+// entry or two, it's fine to just adjust the constant below to match the
+// observed result.
```
cfzimmerman, line 44: Good comment.

Also, how can one inspect the table size after making changes? Not saying it needs to be in the comment, but I'm curious for future reference.

dpd/p4/constants.p4
```diff
@@ -8,22 +8,20 @@ const bit<16> L2_ISOLATED_FLAG = 0x8000;
 
 // TODO: these all need to be bigger. Early experimentation is showing that this
 // is going to need to come either through ATCAM/ALPM or code restructuring.
```
cfzimmerman: Is this TODO still relevant? Np if so, just wanted to check. Looks like the diffs are formatting and cfg only anyway.

dpd/src/counters.rs
```diff
@@ -47,34 +46,22 @@ pub struct Counter {
 }
 
 pub fn get_counter_ids() -> Vec<CounterId> {
```
cfzimmerman: Can this just be a `const COUNTER_IDS: &[CounterId] = &[...];` now?

cfzimmerman (commented):
1. From an aesthetic perspective, this PR is magnificent. Tysm for doing this. Very excited to see it land.
2. What's the plan for validation? All good if E2E testing needs to come at the top of the stack, but do our current CI tests cover these changes well? Both for new functionality and to prevent regressions. If not, I'd appreciate whatever is needed to fill the gap.
3. I'm still pretty new to our P4 program, so I'm unlikely to give a great review on that. Hopefully we can get at least one more reviewer. I'm also experimenting with Claude PR reviews. Imo it's not a replacement for human review, but they are a nice supplement. I'm going to have Claude take a look at this, and I'll comb through some of its comments to see how useful they are.

Lmk once you feel confident about validation, and I'm happy to stamp.

dpd/p4/sidecar.p4
```diff
@@ -2359,6 +2330,8 @@ control Egress(
 				mcast_egress.apply(hdr, meta, eg_intr_md, eg_dprsr_md);
 				mac_rewrite.apply(hdr, eg_intr_md.egress_port);
 			}
+		} else if (meta.bridge_hdr.nat_egress_hit) {
+			egress_filter.apply(meta, eg_intr_md);
 		} else if (eg_intr_md.egress_rid == 0 &&
 		    eg_intr_md.egress_rid_first == 1) {
 			// Drop CPU copies (RID=0) to prevent unwanted packets on port 0
```
cfzimmerman, line 2337: 🤖 This message was written by Claude

**The stated rationale for this reordering looks wrong, and if it is *right* then inbound link-local IPv6 multicast to userspace is broken on hardware. Worth resolving either way.**

The PR body says the NAT egress filter had to be ordered before this check "making sure NAT unicast is not treated as a multicast CPU copy." That is only true if Tofino sets `egress_rid_first == 1` for a non-replicated unicast packet.

But `Services.mcast_inbound_link_local` (sidecar.p4:342-355) produces a packet that takes exactly this path and must *not* be dropped: it sets `meta.is_mcast = true` and `ucast_egress_port = USER_SPACE_SERVICE_PORT`, is never placed in a multicast group (`mcast_grp_a`/`_b` are only set in `MulticastIngress`, which link-local skips), and because `is_mcast` is true it never takes the `else if (!meta.is_mcast)` arm at line 2130 — so it arrives here with `egress_rid == 0`, `nat_egress_hit == false`, and falls straight into this branch.

`test_link_local_multicast_inbound` (`dpd-client/tests/integration_tests/route_ipv6.rs:432`) sends `ff02::1` and asserts it *is* received on `SERVICE_PORT` with a sidecar header. It is `#[ignore]`d, so it is in the `packet-test` set, and `packet-test` is green on this commit. So on tofino-model, `egress_rid_first` is evidently **0** for this packet — which means it is 0 for non-replicated unicast generally, which means NAT unicast was never at risk of hitting this branch and the reordering is defensive rather than load-bearing.

Two possibilities, and only you can settle which:

1. `egress_rid_first` is 0 for unicast on hardware too. Then the reorder is harmless and I would keep it — it reads better — but the commit message should not claim it fixes a drop, because the next person will build a mental model of `egress_rid_first` from that sentence and it will be wrong.
2. `egress_rid_first` is 1 for unicast on real hardware and tofino-model just does not model it faithfully. Then the reorder genuinely is load-bearing for NAT egress — *and* inbound link-local IPv6 multicast to userspace is being silently dropped as `multicast_cpu_copy` on hardware, taking NDP/MLD on the underlay with it, with CI unable to see it.

Case 2 is the one worth ten minutes on hardware before this lands. If it holds, the fix is to exclude `meta.bridge_hdr` service-routed / link-local traffic from this branch too, not just NAT egress.

dpd/p4/sidecar.p4
```diff
@@ -2196,8 +2132,12 @@ control Ingress(
...
 			}
-			meta.bridge_hdr.setInvalid();
-			ig_tm_md.bypass_egress = 1w1;
+			if (meta.nat_egress_hit && !meta.service_routed) {
+				meta.bridge_hdr.nat_egress_hit = true;
+			} else {
+				meta.bridge_hdr.setInvalid();
+				ig_tm_md.bypass_egress = 1w1;
+			}
 		}
```
cfzimmerman, line 2141: 🤖 This message was written by Claude

**A blocked NAT-egress packet is now counted as having successfully egressed.**

Under the old code `egress_filter.apply()` ran up at ingress line 2185 and set `meta.dropped = true`, so the packet took the `if (meta.dropped)` branch and never reached `egress_ctr`. Now the block decision has moved downstream, so ingress cannot know it yet: line 2131 `egress_ctr.count(ig_tm_md.ucast_egress_port)` runs unconditionally for every non-multicast packet, and `mac_rewrite` is applied, before the egress pipeline decides to drop it.

Concretely: a gimlet on port 10 sends Geneve-encapped guest traffic that routes to port 14, which is not registered as an uplink. The operator runs `swadm counters get egress` and sees port 14 incrementing, i.e. the counter asserts these packets left the switch — when every one of them is being dropped in the egress deparser. `Egress` is one of the six always-on counters, so this is the first place someone looks.

This is intrinsic to the move rather than a slip — ingress genuinely no longer has the verdict — so I do not think there is a clean P4 fix short of decrementing, which is not worth it. But it should be stated in the commit message, because `Ingress` vs `Egress` per-port deltas are exactly how you localise a black hole, and this makes that comparison lie for one traffic class. `test_backplane_egress_ipv4_udp` would not notice: it asserts on captured packets, not counters.

dpd/p4/sidecar.p4
```diff
@@ -2383,14 +2356,9 @@ control Egress(
 				underlay_mcast_ctr.count(eg_intr_md.egress_port);
 			}
 		} else {
-			// non-multicast packets should bypass the egress
-			// pipeline, so we would expect this to be 0.
 			unicast_ctr.count(eg_intr_md.egress_port);
```
cfzimmerman, line 2359: 🤖 This message was written by Claude

**Three egress counters change meaning here and none of it is written down — including the comment this hunk deletes.**

The removed lines ("non-multicast packets should bypass the egress pipeline, so we would expect this to be 0") were the only documentation `unicast_ctr` had, and they were accurate. It is now reachable, but only for one narrow population: unicast packets with `bridge_hdr.nat_egress_hit` set that the filter *allowed*. It is not "unicast packets egressing this port," which is what `Unicast` reads as in `swadm counters`. An operator comparing `Unicast` against `Ingress` for a port will see it orders of magnitude lower and reasonably conclude packets are disappearing.

The same move relocates the block itself: `nat_egress_blocked` used to land in `Ingress_Drop_Reason` with the drop attributed to the **ingress** port, and now lands in `Egress_Drop_Reason` (lines 2341-2345) attributed to the **egress** port. So the "which gimlet is misconfigured" signal is replaced by "which non-uplink port did it try to leave by," and anyone with a runbook or saved `swadm` invocation for this drop reason gets a silent zero.

And `common/src/counters.rs:51` still models all of this as `CounterId::Multicast(MulticastCounterId::…)`. That nesting was accurate when the egress pipeline only ran for multicast; now a plain unicast NAT drop reports under a `Multicast(..)` variant, and `MulticastCounterId::Unicast` is a contradiction in terms.

Suggested, in rough order of value:
- replace the deleted comment with what `unicast_ctr` counts now ("NAT-egress-allowed packets only; everything else sets `bypass_egress` and is never seen here");
- note the `Ingress_Drop_*` → `Egress_Drop_*` move for `nat_egress_blocked` in the commit message;
- rename `MulticastCounterId` → `EgressCounterId` (or lift `EgressDropPort`/`EgressDropReason`/`Unicast` out to `CounterId` directly). The `Display` strings are already correct, so `swadm` output and `counters.txt` are unaffected — it is a pure type-honesty change.

dpd/p4/sidecar.p4
```diff
@@ -2383,14 +2356,9 @@ control Egress(
 				underlay_mcast_ctr.count(eg_intr_md.egress_port);
 			}
```
cfzimmerman, line 2357: 🤖 This message was written by Claude

**`underlay_mcast_ctr` is unreachable — `Multicast_Underlay` reads zero forever.**

The preceding arm at line 2351 is `else if (hdr.geneve.isValid())`. This arm's condition begins with the same `hdr.geneve.isValid()`, so it is strictly dominated: anything that could satisfy it was already counted by `external_mcast_ctr` one branch earlier. Underlay multicast is therefore reported as *external* multicast, and `external_mcast_ctr` over-reports by exactly the underlay volume.

This is pre-existing — the lines are identical at the base commit `2475028` and appear here only as context. I am raising it because this PR is what makes it ship: `image.sh` never passed `--multicast`, so in every published artifact to date the entire `Egress` control was `apply { }`. This is the first release where these counters exist on real hardware, and one of the six is wired to nothing.

Swapping the two arms fixes it:

```p4
} else if (hdr.geneve.isValid() &&
           hdr.geneve_opts.oxg_mcast.isValid() &&
           hdr.geneve_opts.oxg_mcast.mcast_tag == MULTICAST_TAG_UNDERLAY) {
	underlay_mcast_ctr.count(eg_intr_md.egress_port);
} else if (hdr.geneve.isValid()) {
	external_mcast_ctr.count(eg_intr_md.egress_port);
}
```

Note this also changes what `Multicast_External` reports, so it is a counter behaviour change rather than a no-op — entirely reasonable to split into its own commit if you would rather not mix it into this one.

dpd-client/tests/integration_tests/table_tests.rs
cfzimmerman, line 44: 🤖 This message was written by Claude

**The three tests that would actually validate `@ways(8)` are the ones CI skips.**

This comment is the PR's argument that `@ways(8)` gets us full occupancy, and `test_routev4_full`/`test_routev6_full` are the evidence. But those only fill sequentially from empty — they say nothing about placement after churn, which is where cuckoo hashing actually fails.

The tests that do cover that are `test_add_target_succeeds_when_table_fragmented_v4`, `..._v4_over_v6` (`route_ipv4.rs:664,716`) and `..._v6` (`route_ipv6.rs:796`) — fill until allocation fails, delete to fragment, assert a new allocation still succeeds. `.github/buildomat/packet-test-common.sh:110` passes `--skip succeeds_when_table_fragmented`, so all three are excluded. That skip is pre-existing and not yours, but it means the green `packet-test` on this commit is not evidence for the property the comment claims.

Failure scenario: `@ways(8)` reaches full occupancy on a fresh table but placement after add/delete churn regresses. A rack that has been up for months and churned routes starts rejecting route-target adds while the table shows free slots, and nothing in CI would have caught it.

At minimum, run those three locally with `@ways(8)` and say in the PR body that they pass. Better, since you are already rewriting this comment: assert a floor rather than an exact count (`assert!(n >= 8000)` plus logging the actual) so placement jitter does not train reviewers to edit the constant, which is also what they would do the day a real regression takes it back to 7164.

dpd/src/mcast/mod.rs
```diff
@@ -2393,15 +2393,16 @@ mod tests {
...
         ];
-        assert_eq!(
-            canonicalize_sources(Some(exact_sources.clone())),
-            Some(exact_sources)
-        );
+        let expected = vec![
+            IpSrc::Exact(IpAddr::V4(Ipv4Addr::new(10, 0, 0, 1))),
+            IpSrc::Exact(IpAddr::V4(Ipv4Addr::new(192, 168, 1, 1))),
+        ];
+        assert_eq!(canonicalize_sources(Some(exact_sources)), Some(expected));
```
cfzimmerman, line 2405: 🤖 This message was written by Claude

**This expectation changed with no corresponding source change, which is worth a line in the commit message.**

`canonicalize_sources` is untouched by this PR and routes through a `BTreeSet`, so it has always deduplicated and sorted — the old assertion was simply never executed. `dpd/src/main.rs` had `#[cfg(feature = "multicast")] mod mcast;`, `test.sh` loops over `tofino_stub tofino_asic softnpu chaos` and never enabled `multicast`, and `multicast-test.sh` only ran `cargo test` in `dpd-client` and `swadm`, never in `dpd`. So the module was cfg'd out of every `cargo test` and every `cargo clippy --deny warnings` the CI runs.

Removing the gate makes 25 unit tests newly live (13 here, 12 in `dpd/src/mcast/validate.rs`) and puts the module under clippy for all four asic backends. `test` is green, so they pass — that is a genuine and undersold benefit of this PR. Worth saying explicitly, though, because a reader six months from now will otherwise read this hunk as "behaviour changed" rather than "the assertion was dead."

dpd/p4/sidecar.p4
```diff
@@ -945,7 +922,14 @@ control RouterLookupIndex6(
 	 * test to pass.  We want the lookup and forward tables to have the same
 	 * capacity from dpd's perspective, and the "default" entry consumes a
 	 * slot in the lookup table.
+     *
+     * Note: we annotate @ways here, increasing cuckoo placement choices to 8
+     * candidate buckets per key. Full occupancy of the route-target table
+     * is a measured result for this build, not a compiler guarantee.
+     *
+     * See https://github.com/p4lang/p4c/blob/a19f1c3d85a867a6288fd983f7bad505ac47d728/backends/tofino/bf-p4c/common/pragma/pragmas.cpp#L1172-L1183.
 	 */
```
cfzimmerman: 🤖 This message was written by Claude

Nit: these added comment lines are space-indented inside a tab-indented block, so the `*` column does not line up with the `/*` above or the `*/` below. Same in `RouterLookupIndex4` at lines 1054-1059.

cfzimmerman (commented):
🤖 This message was written by Claude

Reviewed the full diff at `7790240` (2 commits, 38 files), plus the post-change files at that OID and the real consumers. All five checks are green on this commit, which retires most of the compile-time and capacity speculation a reader might otherwise have.

**The feature removal is complete.** I grepped the whole tree at head for `feature = "multicast"`, `multicast = []`, `features = [..multicast..]`, `-D MULTICAST`, `--multicast` and `ifdef MULTICAST` across `*.rs`, `*.toml`, `*.sh`, `*.p4`, `*.md` — zero hits. The Cargo feature is gone from all six crates that declared it, with no dangling `default = ["multicast"]` and no cross-crate reference left behind. The `require_multicast!`/`not_implemented` unwind in `api_server.rs` is purely mechanical: no path, operation id, parameter, or response type moved, and the endpoints were always declared in the `DpdApi` trait regardless of feature, so the checked-in OpenAPI spec correctly needs no regeneration. `mcast_scope` was deleted from `Filter` along with its `#ifdef` and had no remaining uses. Deleting `multicast-test.sh` loses nothing: both jobs sourced `packet-test-common.sh` and ran the same `cargo test ... --ignored`, so the surviving `packet-test` is a strict superset now that `mod mcast` is unconditional — and it passed here, so the mcast and unicast suites do coexist against one shared dpd/tofino-model.

**The `EgressFilter` move looks correct, and I went at it hard.** The bridged bit is set on exactly the population the old ingress guard covered — the new `else if (!meta.is_mcast)` arm implies `!meta.dropped`, and the inner `meta.nat_egress_hit && !meta.service_routed` completes the old `!meta.dropped && meta.nat_egress_hit && !meta.is_mcast && !meta.service_routed`. The `!meta.is_mcast` half is load-bearing rather than incidental: `MulticastIngress` really does set `meta.nat_egress_hit = true` at sidecar.p4:1781, and the guard is what keeps that off the bridge header. The bit is initialised to `false` for every packet in `IngressParser`, so it is never read undefined; every packet reaching egress still carries a bridge header (the only path that invalidates it also sets `bypass_egress`), so dropping the `setInvalid()`/`setValid()` pair in `EgressParser` is safe because header assignment copies validity; and `guest_traffic_not_allowed` reaches a real drop via `meta.drop_reason != 0` setting `eg_dprsr_md.drop_ctl`. Pruning outer `parse_icmp`/`parse_tcp` from the egress parser is safe — every `hdr.tcp`/`hdr.icmp` reference in `sidecar.p4` is in an ingress control or `IngressDeparser`, and the lone egress one (`hdr.tcp.setInvalid()` at 1955) is a no-op. Bridge header size is unchanged at 16 bits.

I had a draft finding saying the relocated filter had no test coverage. That was **wrong** and I am glad I checked: `dpd-client/tests/integration_tests/nat.rs` already covers both directions. `test_egress_ipv4_{udp,tcp,icmp}` and `test_egress_ipv6_*` run with `uplink_port_registered: true` and assert the packet arrives; `test_backplane_egress_ipv4_udp` and `test_backplane_egress_ipv6_tcp` run with `false` and assert `expected = Vec::new()`, i.e. nothing arrives. Those are `#[ignore]`d and therefore in the `packet-test` set, which is green — so the relocated filter demonstrably blocks and permits correctly. That is real end-to-end coverage of the new bridged-bit mechanism.

**Also checked clean:** the table-name move `Ingress.egress_filter.egress_filter` → `Egress.egress_filter.egress_filter` is updated in all three places (`common/src/table.rs` Display and `TryFrom<&str>`, `asic/src/tofino_common/mod.rs` bfrt path), and the `ucast_egress_port` → `egress_port` key rename matches the convention already used by `mcast_egress.rs`. `counters.txt` is byte-identical to the deleted `counters_multicast.txt` and matches `get_counter_ids()` in order. Table sizing is consistent end to end: P4 `IPV{4,6}_LPM_SIZE` = 8192, `route` tables `SIZE - 1` = 8191, `lookup` tables `SIZE + 1` = 8193, Rust expects 8191 for both; the new `MCAST_SOURCE_FILTER_IPV{4,6}_SIZE` = 1024 are the same values the source-filter tables already had under the old names, so that is a rename with no behaviour change.

**One behaviour change worth a sentence somewhere, too small to anchor:** on softnpu builds the multicast endpoints used to return a clean `501 Not Implemented`; they now attempt the operation and surface `AsicError::OperationUnsupported`. Both are failures and omicron's simulated switch presumably is not driving multicast, so I do not think it matters — but 501 was the more honest answer, and it is a silent API-behaviour change on one backend.

What I did not do: I did not compile the P4 or run anything myself. Everything above is from reading plus the green CI on this commit. The findings below are, with one exception, about observability and rationale rather than the dataplane logic.

**Commit 1 is unambiguously right and I would land it on its own merits.** The `multicast` feature was not a shape, it was a *concept* — a second configuration axis threaded through six `Cargo.toml`s, an `#ifdef` axis in the P4, a macro that made half the API return 501 depending on how you built it, two near-identical buildomat jobs, two `swadm` expect files, and cfg-gated constants in the table tests. Every reader had to hold "which multicast-ness is this binary?" in their head, and the answer was load-bearing. Deleting it pays down a recurring tax; this is exactly the churn worth doing. The evidence that it was worth doing is in the diff: a unit test in `dpd/src/mcast/mod.rs` had been asserting something false for who knows how long, because nothing in CI ever compiled it.

**Commit 2 is where I want to push.** It buys back the stage commit 1 cost by spending an invariant. Before, "guest NAT traffic may only leave on an uplink" was one table applied at one place in one pipeline. Now it is a contract across three sites: ingress sets `bridge_hdr.nat_egress_hit`, ingress must *decline* to set `bypass_egress` on exactly those packets, and the egress parser branches on the bit — and that third site also changes which headers exist in egress at all. The `nat.rs` tests do cover the mechanism end to end, which is the main thing, and it lowers my concern a lot. What remains is the conceptual cost: a reader now has to know that the ingress/egress boundary carries policy state, and the counter story below is the visible symptom of that.

The PR quotes Nils' XXX comment and takes the first half of his suggestion (move it to egress) while deferring the second half (merge it with `MacRewrite`) to a follow-up. Those read to me as *alternatives*, not a sequence. Merging with `MacRewrite` keeps everything in ingress: both are per-port tables keyed on the egress port, both run after routing, and collapsing them removes a table rather than adding a cross-pipeline bit. Nils' objection was "weird semantics to the control plane daemon" — but the daemon already programs two separate port tables here (`uplink_ports` on ingress, `egress_filter`), and `dpd/src/table/uplink.rs` already keeps them in lockstep with rollback on failure, so the daemon-side cost looks like renaming an action, not a new concept. **I have not verified that merging them actually saves a stage.** That is the thing that would change my mind and it is cheap for you to check. If it does, it is the version I would ship, because it removes a table instead of adding an invariant that spans pipelines.

Third: `@ways(8)` and the exact `8191` assertion. Your own comment is honest that full occupancy is "a measured result for this build, not a compiler guarantee," which is precisely the problem — the test now asserts an exact number nobody controls, produced by a compiler placement heuristic. The predictable failure mode is that a future P4 change moves it to 8189, a reviewer "just adjusts the constant to match" per the comment's own advice, and the same reflex fires the day a real regression takes it to 7164. Assert a floor and log the actual. One line, strictly more robust.

**The uncomfortable question.** Both commits point the same way: the pipeline is full, and we are now buying stages by spending invariants — a bridged bit here, a placement pragma there, each individually defensible. #208 already sketched the restructuring that would create real headroom, and it keeps getting deferred because we want to be "efficient and targeted." That instinct is usually right and I would normally be the one arguing for it. But the third bespoke stage-scavenging trick is where targeted stops being cheaper than structural, and nothing in this PR names the plan for the next feature that needs a stage. What is it?

dpd/p4/sidecar.p4
cfzimmerman: This is a real mind bender to read. (2) is a pretty bold claim, so I'm assuming reality is (1). I think claude is saying to reconsider wording in the commit message.

dpd/p4/sidecar.p4
cfzimmerman: The bolded claim is self explanatory. Seems reordering might interact poorly with NAT egress counters, and that warrants a second look.

dpd/p4/sidecar.p4
cfzimmerman: I'm generally skeptical when claude says more comments are needed, so that's up to you.

But the interaction with multicast counter types on the rust side seems worth consdering.

dpd/p4/sidecar.p4
cfzimmerman: Is this resolved in this PR? https://github.com/oxidecomputer/dendrite/pull/362

dpd-client/tests/integration_tests/table_tests.rs
cfzimmerman: I think tl;dr: can we un `--skip` these tests in CI, or are they problematic for other reasons? If not, maybe make sure they pass?

dpd/src/mcast/mod.rs
cfzimmerman: I think this is fine.

dpd/p4/sidecar.p4
cfzimmerman: Lol do this if you want, but imo there are bigger fish to fry than tabs v. spaces.

cfzimmerman:
Please lmk if you find the extra claude reviews useful or annoying. Some of the comments are really unreadable, so I tried to at least preread a bit.

dpd/p4/sidecar.p4
zeeshanlakhani: I'll line it up (I use Emacs with spaces).

dpd-client/tests/integration_tests/table_tests.rs
zeeshanlakhani: This has been through codegen and running the `table_tests`. Otherwise, `swadm table list` / `swadm table dump <name>` at runtime. Although swadm doesn't display capacity, the dump response carries the size (max # entries) if you want the number.

dpd/p4/constants.p4
zeeshanlakhani: Oh, I didn't touch this, but I guess it could go away? Removed.

dpd/p4/sidecar.p4
zeeshanlakhani: `egress_rid_first` is PRE-populated (Packet Replication Engine). Regular unicast never engages with the PRE, so it arrives `0/0` (`egress_rid`/`egress_rid_first`) and definitely can't match the CPU-copy arm. 

I'll keep the reordering (and I removed the line from the description).

dpd/p4/sidecar.p4
zeeshanlakhani: Yep, it's resolved in the follow-up. It made sense to separate out (and keep it so).

dpd/p4/sidecar.p4
zeeshanlakhani: The goal here is stage saving. No doubt that a blocked NAT-egress packet now increments the ingress `egress_ctr` before the egress pipeline itself drops it. Preserving the old meaning would require separate ingress/egress counters and aggregation, and I'm not sure that's worth it, as the current semantic meaning is not wrong; it is being assigned to the egress port.

dpd/p4/sidecar.p4
zeeshanlakhani: The counter observation is indeed correct. 

A blocked NAT-egress packet increments the ingress `egress_ctr` before the egress pipeline drops it (the drop itself is tracked separately). The unicast counter does count allowed NAT-egress packets only, as ordinary unicast bypasses the egress pipeline and never reaches it.


`nat_egress_blocked` is now recorded under `Egress_Drop_*`, which is attributed to the attempted egress port. The stale comment on `unicast_ctr` was removed. I've also renamed the types/variants on the Rust side: `MulticastCounterId` is now `EgressCounterId`, `CounterId::Multicast(..)` -> `CounterId::EgressPipeline(..)`, and the `Egress*` repetition inside the enum is gone: `EgressCounterId::{DropPort,DropReason}`. Display strings are unchanged, so swadm output and the API are untouched.

dpd-client/tests/integration_tests/table_tests.rs
zeeshanlakhani: This predates any work here ~ #322, but let me see if they do pass.

cfzimmerman (approved):
Recent changes lgtm. Although a second reviewer warranted before merge, particularly to review the p4 diffs.

dpd/p4/constants.p4
```diff
@@ -6,24 +6,20 @@
 
 const bit<16> L2_ISOLATED_FLAG = 0x8000;
 
-// TODO: these all need to be bigger. Early experimentation is showing that this
-// is going to need to come either through ATCAM/ALPM or code restructuring.
```
FelixMcFelix, line 10: I think this comment still holds true. 1024 entries for, e.g., NAT is quite paltry compared to how many VMs you can feasibly run on a single rack (particularly with each OPTE port consuming several such entries).

asic/src/tofino_asic/mod.rs
```diff
@@ -395,28 +391,18 @@ impl Handle {
...
         let rt = tofino_common::BfRt::init(&p4_dir)?;
         let phys_ports = ports::init(dev_id)?;
         let eth_connector_id = phys_ports.eth_connector_id;
 
         // Note: we assume that bf_mc_init() has been called as part of the
         // bf_switch_init() operation.
-        #[cfg(feature = "multicast")]
         {
             bf.mcast_hdl = mcast::create_session()?;
         }
```
FelixMcFelix, line 408: Is the separate lexical block for this statement required?

dpd/src/link.rs
FelixMcFelix, line 1634: Nit: a few more scoped blocks leftover in this file.

dpd/src/macaddrs.rs
```diff
@@ -434,7 +433,6 @@ impl Switch {
 
         // Reset ingress and egress MAC tables and Port ID table(s).
         mac::reset(self)?;
-        #[cfg(feature = "multicast")]
         {
             mac::mcast_reset(self)?;
             mcast::mcast_egress::reset_bitmap_table(self)?;
```
FelixMcFelix, line 438: Ditto.

dpd/p4/sidecar.p4
```diff
@@ -945,7 +922,14 @@ control RouterLookupIndex6(
...
 	 * capacity from dpd's perspective, and the "default" entry consumes a
 	 * slot in the lookup table.
+	 *
+	 * Note: we annotate @ways here, increasing cuckoo placement choices to 8
+	 * candidate buckets per key. Full occupancy of the route-target table
+	 * is a measured result for this build, not a compiler guarantee.
+	 *
+	 * See https://github.com/p4lang/p4c/blob/a19f1c3d85a867a6288fd983f7bad505ac47d728/backends/tofino/bf-p4c/common/pragma/pragmas.cpp#L1172-L1183.
 	 */
+	@ways(8)
```
FelixMcFelix, line 932: Out of curiosity, do we have any idea what this defaults to?

dpd/p4/sidecar.p4
FelixMcFelix: I think the change in meanings of counters is confusing enough that it should be documented or worked around.

My current read is that to get the true egress count on a port, you need to compute `pipe.Ingress.egress_ctr[port_idx] - pipe.Egress.drop_port_ctr[port_idx]`. Not unreasonable, but it requires decent knowledge of the dataplane to make sense. Thinking on alternate approaches:

* Could we move the egress counter hit into the else block? This way the sum of packets allowed to leave on a port is `pipe.Ingress.egress_ctr[port_idx] + pipe.Egress.unicast_ctr[port_idx]` (plus ` + pipe.Egress.mcast_ctr[port_idx]`, if you're so inclined). The counter names are still a little misleading this way, in that `Ingress.egress_ctr` would be underlay plus inbound nat'd traffic, whereas `Egress.unicast_ctr` is outbound traffic being decapsulated successfully.
* Does `bypass_egress` give us meaningful latency improvement? Does it make enough difference to prevent us pushing _all_ packets through the egress pipeline, and then we only need `pipe.Egress.unicast_ctr[port_idx]`? This would also mesh well with making `MacRewrite` an unconditional part of the egress pipeline in future.

FelixMcFelix (commented):
Thanks Zeeshan.

dpd/p4/constants.p4
zeeshanlakhani: Can return/extend it.

dpd/src/link.rs
zeeshanlakhani: Yeah, I moved some things around, will clean it up.

dpd/p4/sidecar.p4
zeeshanlakhani: 4. I'll add that to the comment.

zeeshanlakhani:
On https://github.com/oxidecomputer/dendrite/pull/371#discussion_r4007995518,

I was being a bit conservative here, but the second alternative approach is what I already explored in #208 (with the MAC rewriting, however). I'll make that update, as it's cleaner anyway and sets us up for that move.

FelixMcFelix (approved):
Okay, I'm happier with these changes now. I think this keeps the counter state somewhat sensible. I don't know what we want the wider testing plan to be yet before we merge, given the risk of London-bridges type scenarios. Are we wanting to validate the whole PR stack with product assurance etc. beforehand?

zeeshanlakhani:
> Okay, I'm happier with these changes now. I think this keeps the counter state somewhat sensible. I don't know what we want the wider testing plan to be yet before we merge, given the risk of London-bridges type scenarios. Are we wanting to validate the whole PR stack with product assurance etc. beforehand?

Good question on the latter. I'll tag @askfongjojo and @sion42x here as well.

PR #374, opened 2026-09-14
Fix indefinite port_settings_* failures

### LINKED ISSUE #354: DPD partial failures can be fatal
# Summary

"fatal" feels like the wrong word because I associate it with panic. But basically the rng txn chaos test sometimes gets dpd into a state where a repeated `port_settings_*` operation will always fail due to soft state / SDE mismatches. This pointed me toward a broader pattern of interrupted writes causing unrecoverable states. The general solution is replacing cases of "success if the desired state transition occurred" with "success if the final state matches the request". The latter behaves batter among reconcilers and absorbs previously-interrupted writes.

### Background

`test_port_settings_txn_sweep` fails about 1/10 runs locally. I believe I've also seen this test fail in CI. The failures are a class of bugs related to partial config creating an unexpected state for future config. Reproduced in the tests on this branch: accounts.google.com/gsi/transform

Here's an example of one such failure:
- port_settings_clear
- unplumb_link
    - uplink_clear
    - clear_mac_config
        - table op fails due to chaos
        - unplumb exits in error
- unplumb_link (tries again)
    - uplink_clear
        - table op fails because the uplink is already cleared

In this case, the uplink is already in the state we want, but the code errs because the expected state transition didn't occur. And now we'll get this failure perpetually.

The PR identifies and fixes some more cases. But the issue is that various create/delete actions expect state transitions, and reconciling through partial failures sometimes loses a state transition to some other error. So the state doesn't transition during the next reconciliation attempt, and that's interpreted as an error indefinitely.

### port_settings_apply

Here's an example from the repro test failing indefinitely at `port_settings_apply`.

```
{"msg":"failed to add ipv4 address","v":0,"name":"dpd","level":50,"time":"2026-09-05T13:41:36.242997652Z","hostname":"helios","pid":22379,"error":"ASIC error: Exists","port":33,"addr":"211.33.123.210"}
{"msg":"failed to add ipv4 address","v":0,"name":"dpd","level":50,"time":"2026-09-05T13:41:36.747271693Z","hostname":"helios","pid":22379,"error":"ASIC error: Exists","port":33,"addr":"211.33.123.210"}
{"msg":"failed to add ipv4 address","v":0,"name":"dpd","level":50,"time":"2026-09-05T13:41:37.250982862Z","hostname":"helios","pid":22379,"error":"ASIC error: Exists","port":33,"addr":"211.33.123.210"}
cory@helios:~/dev/dendrite/dpd-client$ cat /tmp/settings_eventually_reconcile-dpd.stdout | rg "ASIC error: Exists" | wc -l
     124
cory@helios:~/dev/dendrite/dpd-client$
```

### DPD reconciler

A partial write to the mac tables can cause the same. This is from another test.

```
{"msg":"Failed to program mac and port mapping: Switch(Exists)","v":0,"name":"dpd","level":50,"time":"2026-09-06T11:20:38.839686537Z","hostname":"helios","pid":4419,"link":"qsfp0/0","unit":"reconciler"}
{"msg":"failed to add entry to PortMacAddress: Switch(Exists)","v":0,"name":"dpd","level":20,"time":"2026-09-06T11:20:39.090990028Z","hostname":"helios","pid":4419}
{"msg":"set mac on 33 in table Ingress.mac_rewrite.mac_rewrite: a8:40:25:4a:1c:c2 failed: Switch(Exists)","v":0,"name":"dpd","level":50,"time":"2026-09-06T11:20:39.091017025Z","hostname":"helios","pid":4419}
{"msg":"Failed to program mac and port mapping: Switch(Exists)","v":0,"name":"dpd","level":50,"time":"2026-09-06T11:20:39.0910418Z","hostname":"helios","pid":4419,"link":"qsfp0/0","unit":"reconciler"}
```

cfzimmerman:
### Prefix parlance

- "create" fails if the target entry already exists. Transition from nothing to something.
- "delete" fails if the target entry does not exist. Transition from something to nothing.
- "set" succeeds as long as the target entry ends up in the collection. No issue if the target already exists.
- "clear" succeeds as long as the collection ends without the entry. No issue if the entry already does not exist.

In my view, set and clear are generally more useful APIs for reconcilers.

# Summary

- Table operations previously used create/delete, which caused unrecoverable failures in sequences of the following form. I propose using set/clear instead to absorb these failures in subsequent reconciliation attempts.
    - Example 1: `reconcile_link`, which doesn't use txn/rollback:
        - Attempt 1:
            - delete asset 1. Success.
            - delete asset 2. Fails. Return error.
        - Every subsequent attempt:
            - delete asset 1. Fails because asset 1 doesn't exist. 
    - Example 2: `port_settings_apply`, which uses txn/rollback:
        - Attempt 1:
            - create asset 1. Success.
            - create asset 2. Fails.
            - Roll back asset 1. Fails. Unlikely but possible. This happens in the txn chaos test.
        - Every subsequent attempt:
            - create asset 1. Fails because asset 1 already exists.
- Changing that raised an issue with how addresses are globally synchronized. We're currently using `AsicError::Exists` from the SDE to prevent one link from stealing an address belonging to another. set/clear is actually a regression in that it ignores previous ownership. However, afaict DPD already carries a grave distrust of the SDE, and using it for multithreaded synchronization seems questionable. Since link/loopback addresses share global switch tables, this PR proposes tracking these addresses in a global table and not relying on the SDE for synchronization.
- There was substantial code duplication in the `link` and `port_settings` modules for IPv4 and IPv6  operations. The new global table uses generics, which helped deduplicate some code.

# Fixes

Running only the first commit yields the following test failures. `test_port_settings_txn_sweep` fails roughly 1/10 runs, which motivated this PR in the first place. The same failure case is reproduced in `settings_eventually_reconcile`.

```txt
failures:
    chaos_tests::port_settings::deletion_doesnt_leak_table_entries
    chaos_tests::port_settings::link_addrs_are_isolated
    chaos_tests::port_settings::link_init_recovers
    chaos_tests::port_settings::links_cannot_steal_addrs
    chaos_tests::port_settings::loopback_addrs_are_isolated
    chaos_tests::port_settings::partial_failures_are_recoverable
    chaos_tests::port_settings::settings_eventually_reconcile

test result: FAILED. 15 passed; 7 failed; 0 ignored; 0 measured; 0 filtered out; finished in 90.01s
```


dpd-client/tests/chaos_tests/port_settings.rs
cfzimmerman, line 1: This has a lot of new tests. But they're are hand written, and I think they'll be useful at catching regressions. I don't take a political position on TDD, but I did kind of use TDD in this PR when isolating bugs.

There's a decent bit of rng, but they should be stable in CI.

cfzimmerman (commented):
🤖 This message was written by CLU

Reviewed at `05ff476` (draft). The idempotency half of this is convincing: I traced each converted table op and `table_entry_set`/`table_entry_clear` do have the semantics the PR body claims, on all four backends (tofino maps `BF_ALREADY_EXISTS`/`BF_OBJECT_NOT_FOUND` to `Exists`/`Missing`, chaos checks its key set before rolling the dice, softnpu is unconditionally `Ok`). Both wedges the body describes are addressed at the right layer.

Things I checked and am not raising, so you don't re-investigate them:

- **The reusable-owner-name orphan is fixed structurally, not by convention.** `LinkMap::delete_link` is now the only path that removes an entry from `LinkMap.0`, it purges the owner's addresses first, and it `?`s so the slot survives a failed purge. All three callers inherit that. `replace_link` keeps the slot and passes the same `Arc` and owner key, so the revive path inherits the previous incarnation's addresses deliberately rather than orphaning them — which is also a quiet improvement over the old `Link::new` that dropped them and leaked the ASIC rows.
- **The MAC is not double-freed or leaked.** `add_link` winds the `free_mac_address` closure first, so it unwinds last — after the `delete_link` closure — and the `ctx.link(link_id).is_err()` guard means a link that survives rollback keeps its MAC until the reconciler deletes it.
- **Lock ordering is consistent:** `switch.links` -> `Link` -> `AddrMap` -> table, with `AddrMap` always innermost. `handle_port_update` is driven off an unbounded mpsc, so the new chaos `port_enable_set` send cannot deadlock against the link lock the reconciler holds.
- **No TOCTOU** between the ownership check and the ASIC write: `try_set`/`try_clear` do both under one `&mut self`.
- The body's list of 7 tests that fail on commit 1 alone looks right; I traced why `deletion_prevails` converges under the old non-idempotent deletes and so is correctly absent from that list. I did not build or run anything.

The customer problem is real and worth the trip: one transient SDE error leaves a port permanently unmanageable until `dpd` restarts. That is the kind of defect that turns into a support case, and per the reproductions it is on `main` today. The idempotency change is the smallest thing that fixes it, and if this PR were only `table_entry_set`/`table_entry_clear` plus the `plumbed`-flag paths that use them, I would argue for landing it as-is and fast.

The part that deserves pressure is the global `AddrMap`, because it is not required by the wedge — it is required by a *consequence* of the fix. Dropping `AsicError::Exists` as the cross-owner guard removed the only thing stopping a link address and a loopback address with the same IP from overwriting each other, and for IPv6 that really is the same key (`match_keys_ipv6`'s drop key is `{val: 0, mask: 0}`, byte-identical to the loopback claim key). So the map exists to re-supply a guarantee the SDE used to provide. That is a legitimate reason, and the body says so. Three costs are worth naming out loud before this lands:

1. **It is a new single point of failure with an abort policy.** Every link's and loopback's address bookkeeping now lives behind one `RwLock<AddrMap>` whose internal invariants are enforced with `assert!`/`expect`/`panic!`, three of them on read paths, under `panic = "abort"` in both profiles. Before, a bookkeeping bug corrupted one `Link`; now it takes the daemon down during a `GET`.

2. **It collapses concurrency that existed.** `PortAddrIpv4` and `PortAddrIpv6` are separate P4 tables behind separate `Mutex<Table>`es, and address state used to live per-`Link` behind the link mutex — so v4 and v6 address work on different ports proceeded in parallel. Now every address operation on the switch serializes on one write lock that is *held across the SDE call*. On a 32-port sidecar coming up, that is the whole switch's address plumbing in one queue. I have not measured it and it may be noise next to the rest of link bring-up, but it is a change worth knowing you made rather than discovering.

3. **"Exactly one owner across loopback and all links" is a new, stricter API contract.** On `main`, two links on different ports claiming the same IP failed messily via `Exists`; now it is a clean 409. Cleaner — but it is a behaviour change omicron's reconciler can observe, and it is not in the body's summary of what changed. One line there would help whoever reads this in six months.

Which points at the scoping question rather than the design: this PR bundles three separable changes — idempotent table ops, global address ownership, and the v4/v6 generic dedup — and only the first is needed for the reported failure. Split, the idempotency fix lands on its own merits immediately, and the ownership model gets its own review where the abort policy, the lock granularity, and the new 409 contract are argued directly instead of riding along with a bug fix nobody wants to hold up. The dedup is the easiest of the three to defend and the easiest to defer.

None of that is an argument against the shape you've chosen for `AddrMap` — having read the alternatives, keeping the soft state and keying it by owner is more honest about the hardware than what `main` does, and the collision really does have to be caught above the table layer for v4 (where the claim/drop keys differ in dpd's own key set but collide in the TCAM). It is an argument that the second and third changes should have to earn their way in separately from the first.

**Findings:** 2 possible bug, 4 conceptual error, 1 unclear assumption, 2 test gap, 2 maintainability, 1 nit

dpd/src/link.rs
```diff
@@ -726,65 +768,21 @@ impl Switch {
...
-                let to_delete = std::mem::take(&mut link.ipv6).into_keys();
-                port_ip::ipv6_delete_many(self, link.asic_port_id, to_delete)?;
-            }
+            link_lock
+                .lock()
+                .unwrap()
+                .clear_link_addresses::<IpAddr>(self, tag)?;
         }
         Ok(())
     }
```
cfzimmerman, line 784: 🤖 This message was written by CLU

**Category: Possible bug**

The `?` inside this loop is a behaviour change for both of this function's callers, and it makes retries useless.

Before, the `Some(tag)` case went through `clear_link_addresses_locked`, which did `let _ = port_ip::ipv4_delete_many(...)` and so processed *every* link regardless of ASIC errors; the `None` case went through `clear_link_state`, whose `ipv4_delete_many`/`ipv6_delete_many` also always returned `Ok`. Neither could fail. Now one failing link aborts the sweep.

Failure scenario: one permanently stuck `PortAddrIpv4` row on `qsfp0/0` — exactly the state `deletion_doesnt_leak_table_entries` constructs. `DELETE /all-settings/{tag}` (`api_server.rs:1745`) and `POST /all-settings` (`api_server.rs:1769`) then return an error having cleared nothing on `qsfp1`..`qsfp31`, and because `LinkMap.0` is a `BTreeMap` iterated in a fixed key order, every retry stops at the same link. Previously the caller got a 204 and full clearing everywhere except the stuck row.

Not lying about the stuck row is the right instinct and I am not asking you to go back to swallowing errors. But refusing to service the other 31 ports is a new denial, and it is inconsistent with `reset_all` itself, which collects an error per subsystem and keeps going. Collecting the errors and continuing (returning the last one, as `Rollback::unwind` does) would give you both.

Also on line 771: `siwtch`.

asic/src/chaos/mod.rs
```diff
@@ -494,6 +497,14 @@ impl AsicOps for Handle {
         unfurl!(self, port_enable_set);
         let mut ports = self.ports.lock().unwrap();
         get_port_mut(&mut ports, port_hdl)?.enabled = val;
+
+        if let Some(chan) = self.updates.get() {
+            let _ = chan.send(PortUpdate::Enable {
+                asic_port_id: port_hdl.connector.as_u16(),
+                enabled: val,
+            });
```
cfzimmerman, line 505: 🤖 This message was written by CLU

**Category: Possible bug**

This computes an `asic_port_id` by a different rule than the handle's own `port_to_asic_id` twenty lines below. `Connector::as_u16()` (`aal/src/ports.rs:135`) returns `0` for `Connector::CPU`, while `Handle::port_to_asic_id` returns `CPU_PORT == 47`. They agree for QSFPs only by coincidence (both are `n`).

Failure scenario: `rev_chaos_map` maps `PortId::Internal(0)` to `Connector::CPU`, so a chaos test that enables a link on the internal port emits `PortUpdate::Enable { asic_port_id: 0 }`. `get_connectors` numbers QSFPs `1..=radix`, so asic id 0 names no port — `handle_port_update` fails `asic_id_to_port_link(0)` and logs, and `plumbed.enabled` is never set for that link. A test polling `link_enabled_get` (which under `chaos` reads `plumbed.enabled`, see the separate comment on `api_server.rs`) would then hang until `LONG_ENOUGH`, with the reason buried in dpd's log.

I could not make this fire today: the chaos harness passes no `--mac-base` and no `--port-config`, and the non-`tofino_asic` `set_base_mac_address` invents a random MAC without creating a CPU link, so nothing enables an internal-port link. It is latent rather than live. `self.port_to_asic_id(port_hdl)` is the one-line fix, and is what any future `PortUpdate::LinkUp`/`FSM` sender here would need too.

dpd/src/addr.rs
```diff
@@ -0,0 +1,387 @@
...
+                tag: spec.tag.clone(),
+            });
+        }
+
+        if owner_filter != &spec.owner {
+            return Err(DpdError::AddrOwnerConflict {
+                addr,
+                owner: spec.owner,
+            });
+        }
```
cfzimmerman, line 311: 🤖 This message was written by CLU

**Category: Conceptual error**

The tag is checked before the owner, so a cross-owner conflict is reported as a tag conflict and points the operator at a fix that cannot work.

Failure scenario: `X` is owned by `Link { qsfp0, 0 }` under tag `A`. `port_settings_apply(qsfp1, tag="B", [X])` returns 409 `"Address X exists, but it is owned by another tag: A"`. The natural next move — reapply under tag `A` — produces a *different* 409 naming a different problem (`AddrOwnerConflict`), because the tag was never the obstacle. The owner invariant is the stronger one; the tag comparison only means anything once the owner matches, so the two checks should be in the opposite order.

This is masked by the PR's own tests. `partial_failures_are_recoverable:1023` (`expect_err("Stuck entry cannot be stolen by another tag.")`) uses `TAG2` on `qsfp1`, so it believes it is asserting `AddrOwnerConflict` but actually receives `AddrTagConflict`; `links_cannot_steal_addrs` uses `TAG1` on both ports and does get `AddrOwnerConflict`. Nothing distinguishes the two orderings.

dpd/src/table/mac.rs
```diff
@@ -37,7 +37,7 @@ fn mac_set_common(
     let match_key = MacMatchKey { port };
     let action_data = MacAction::Rewrite { mac };
 
-    match s.table_entry_add(type_, &match_key, &action_data) {
+    match s.table_entry_set(type_, &match_key, &action_data) {
```
cfzimmerman, line 40: 🤖 This message was written by CLU

**Category: Conceptual error**

Switching to `table_entry_set` here makes the doc comments on both public entry points state the opposite of what the code now does. `mac_set` (line 70) and `mcast_mac_set` (line 102) both say:

```
/// Add a new entry to the MAC table.
///
/// An error is returned if the entry already exists. Use `mac_update` instead.
```

An existing entry is now overwritten, not rejected — which is the whole point of the change. And `mac_update` does not exist anywhere in `dpd/src`; those two comments are its only mentions.

That sentence is exactly the kind of claim the next person reading this table will believe, and the previous `Exists`-forever wedge is what happens when someone acts on it. Worth rewriting both to say the entry is rewritten if present, since that is now the contract the reconciler depends on.

dpd/src/api_server.rs
```diff
@@ -904,8 +906,19 @@ impl DpdApi for DpdApiImpl {
...
+
+        #[cfg(not(feature = "chaos"))]
+        let src = link::Source::Config;
+
+        // Let chaos tests inspect SDE state to detect divergences
+        // with link soft state. I question why this isn't the default
+        // behavior, but flagging this saves that conversation for
+        // another day.
+        #[cfg(feature = "chaos")]
+        let src = link::Source::Switch;
```
cfzimmerman, line 918: 🤖 This message was written by CLU

**Category: Conceptual error**

A `cfg` fork in an HTTP handler's *semantics* is the wrong place to defer this conversation. `link_enabled_get` is a public endpoint; in a chaos build it answers a different question than in a shipped one, and nothing in the OpenAPI document or `swadm link` says so.

Two concrete consequences:

- `link_init_recovers` only tests anything because of this fork. Under `Source::Config` its loop exits after the first `link_enabled_set` and the test asserts nothing — so the only coverage of the `set_port_mapping_entry` idempotency fix runs against an endpoint that behaves differently from the shipped one, and would silently become a tautology if this cfg were ever removed.
- Conversely, the `Source::Switch` arm of the new enum is never exercised in a production build, and `Source::Config` is never exercised in a chaos build, so neither arm is covered where it ships.

If the comment's instinct is right and callers want `plumbed.enabled`, the shapes that don't fork the contract are a query parameter on this endpoint, or a separate `link_plumbed_get` that both builds expose — either way the test and production read the same code path. (`LinkView.enabled` already exists on `link_get`, so a second field there is another option.)

dpd/src/table/port_ip.rs
```diff
@@ -165,27 +165,30 @@ fn endeavour_to_repair(
...
 }
 
-/// Add one IPv4 address to the ASIC tables.
-pub fn ipv4_add(s: &Switch, port: u16, ipv4: Ipv4Addr) -> DpdResult<()> {
-    ipv4_add_work(s, port, ipv4)
+/// Places this IPv4 address in the appropriate ASIC tables.
+///
+/// If the address does not yet exist, an entry is created.
+/// If it does, the entry is updated.
+pub fn ipv4_set(s: &Switch, port: u16, ipv4: Ipv4Addr) -> DpdResult<()> {
```
cfzimmerman, line 190: 🤖 This message was written by CLU

**Category: Conceptual error**

"If it does, the entry is updated" describes a branch that `AddrMap` makes unreachable for this table, and it is the claim most likely to be relied on later.

`try_set` only reaches `set_asic` on the `Entry::Vacant` arm — if soft state already holds the address under the same owner and tag it returns `Ok(false)` without touching the ASIC. Since `addr.rs:81` asserts "an address is in this soft state IFF it's in an asic table", the `Exists` -> `table_entry_update` fallback inside `table_entry_set` is dead for `PortAddrIpv4` and `PortAddrIpv6`. It is genuinely load-bearing for the reconciler-driven tables (`mac_set`, `set_port_mapping_entry`, `uplink_set`), which is what `link_init_recovers` covers — but not here.

The corollary is worth stating in the comment too, because it is the actual recovery story for these two tables: nothing ever re-drives a `port_ip` row that soft state believes exists. Recovery comes from *retaining* the soft state so no later add hits `Exists`, not from rewriting the row. Same applies to `ipv6_set`, which documents itself as the "IPv6 equivalent".

dpd/src/addr.rs
```diff
@@ -0,0 +1,387 @@
...
+                port_ip::ipv6_set(switch, asic_id, *v6)
+            }
+            (IpAddr::V6(v6), AddrOwner::Loopback) => {
+                port_ip::loopback_ipv6_set(switch, *v6)
+            }
+        }
+    }
+
+    /// Removes this entry from the corresponding asic table.
+    ///
```
cfzimmerman, line 346: 🤖 This message was written by CLU

**Category: Unclear assumption**

`set_asic` and `clear_asic` call back into `Switch` to re-derive an ASIC id the caller already has, and the invariant that makes that safe is not the one the `# Synchronization` block states.

Every caller of `try_set`/`try_clear` on a link owner holds a `&mut Link` whose `asic_port_id` came from this same `port_link_to_asic_id` call at construction, so this is a redundant lookup that also adds a new mid-operation failure point per address (under `uniform_set` chaos, `port_to_asic_id` is itself injectable — which is fine for the tests, but it is a new `unfurl!` in the address path).

The part that will bite is the locking. This runs with the `AddrMap` write lock held *and* the link mutex held by the caller. It is safe only because `port_link_to_asic_id` is a pure `port_map` + `asic_hdl` translation that takes no link locks. Its sibling `link_asic_port_id` — one word away, same file, same signature, same return value — goes through `link_fetch`, which takes `switch.links` and then the link mutex. Swapping one for the other self-deadlocks on a non-reentrant `Mutex`, and nothing in the code or the doc says not to: the `# Synchronization` note promises only "will lock asic tables", which is not the constraint that matters here.

Either state it ("these methods must not acquire `switch.links` or a `Link` lock") or remove the hazard by threading the ASIC id in from the caller.

dpd-client/tests/chaos_tests/port_settings.rs
```diff
@@ -883,6 +968,403 @@ async fn apply_fails_on_tag_conflict() -> anyhow::Result<()> {
...
+    let settings = client
+        .port_settings_apply(&port_id, Some(TAG1), &v4_only)
+        .await
+        .context("Apply should succeed because we can at least overwrite the IPv4 table entry")?;
+
+    let registered_v4 = settings
+        .links
+        .values()
+        .any(|link| link.addrs.contains(&addrs.v4_entry.addr.into()));
+    assert!(registered_v4, "Address was reclaimed");
```
cfzimmerman, line 1037: 🤖 This message was written by CLU

**Category: Test gap**

This assertion is satisfied by an empty diff — no table operation runs, so the test does not cover the recovery it names.

State at this point: the link is in the `LinkMap` with `addrs = {v4: TAG1}` and `delete_me = false` (the failed `port_settings_clear` wound a closure that reset it). So `LinkSpec::from(&Link)` equals `LinkSpec::try_from_settings(v4_only, TAG1, Some(&before))` field for field — same `speed`, `fec: Some(PortFec::None)`, `autoneg`, `kr`, `tx_eq: None`, `allow_ddm_traffic`, same `addrs` — and `calculate_links` hits `if before == after { return None }` (`port_settings.rs:260`). `self.links.modify` is empty, `do_execute` does nothing, and the apply returns the cached settings.

So all this proves is that the stuck entry is still *attributed* to the link, which the two preceding `expect_err`s already established. And it could not prove more: `try_set` short-circuits to `Ok(false)` before `set_asic` whenever the address is already in soft state under the same owner and tag, so the ASIC row is never overwritten — which also makes the doc comment's "so it can at least be overwritten in the next port_settings_apply" (lines 974-978) the wrong mechanism. Recovery here is soft-state retention, not overwrite.

To exercise a real re-drive, give the apply something else to do — e.g. request the stuck v4 address *plus* a second address, so the link enters `modify` with a non-empty change set.

Minor, same block: `assert!(registered_v4, "Address was reclaimed")` prints on failure, so a regression logs the success sentence.

dpd-client/tests/chaos_tests/port_settings.rs
```diff
@@ -519,8 +527,90 @@ fn random_port_settings() -> PortSettings {
...
+            .is_err()
+        {
+            *apply_ct += 1;
+            self::slow_down().await;
+        }
+
+        while client.port_settings_clear(&port_id, Some(TAG1)).await.is_err() {
+            *clear_ct += 1;
+            self::slow_down().await;
+        }
```
cfzimmerman, line 600: 🤖 This message was written by CLU

**Category: Test gap**

These loops discard the error, so the failure mode the test exists to catch is the one it reports least about. A genuine wedge here is a *deterministic* error repeating forever, and what you get is `Timed out waiting for successful reconciliation` after 90s with no indication of which of the three calls wedged or on what — the diagnosis has to come from `/tmp/settings_eventually_reconcile-dpd.stdout`.

Keeping the last error per loop and folding it into the timeout context would make the test self-diagnosing, and the retry counters you already print are the natural place to carry it. `deletion_prevails` (line 1233) has the same shape.

dpd/src/addr.rs
```diff
@@ -0,0 +1,387 @@
...
+            .flat_map(|addrs| addrs.iter())
+            .filter_map(move |addr| {
+                let spec = self
+                    .global
+                    .get(addr)
+                    .expect("If an address is in mirror, it must be in global");
+                assert_eq!(&spec.owner, owner, "Address owners must concur");
+                Some((IpAddrLike::from_ip(addr)?, spec.tag.as_str()))
+            })
+    }
```
cfzimmerman, line 214: 🤖 This message was written by CLU

**Category: Maintainability**

These invariant checks abort the daemon from a read-only path. `Cargo.toml` sets `panic = "abort"` in both `[profile.dev]` and `[profile.release]`, so this is not a failed request — it is `dpd` gone, and with it the data plane's control.

`iter_by_owner`'s `.expect(...)` and `assert_eq!` run on `loopback_ipv4_list`, `loopback_ipv6_list`, `Link::settings`, `LinkSpec::from(&Link)` and `link_local()`; `owner_addr_range`'s `.expect(...)` (line 283) runs on `link_ipv4_list`/`link_ipv6_list`. So every `port_settings_get` and every address listing is a place a `global`/`mirror` divergence becomes an outage.

I could not find a divergence path — every mutation goes through `try_set`/`try_clear`/`try_retain_by_owner` and each maintains both maps — so this is about blast radius, not a live bug. But `mirror` is pure derived state: it can be rebuilt from `global` by definition, and the comment above the struct says as much. `debug_assert!` on the read paths, with the release build either trusting `global` or logging the divergence, keeps the tests as strict as they are now without making a `GET` able to take down the switch. That the previous incarnation of this map made the same choice is a reason to revisit it, not to keep it.

dpd/src/loopback.rs
```diff
@@ -4,121 +4,43 @@
...
-}
-
-/// Initialize loopback data to empty sets of IPv4 and IPv6 addresses.
-pub fn init() -> LoopbackData {
-    LoopbackData { v4_addrs: BTreeSet::new(), v6_addrs: BTreeSet::new() }
+#[derive(Default)]
+pub struct Loopback {
+    /// The global address tracker. Links should clone and use this too.
+    pub addrs: Arc<RwLock<AddrMap>>,
 }
```
cfzimmerman, line 17: 🤖 This message was written by CLU

**Category: Maintainability**

This is now the switch-wide address map for every link, and the type it lives on says "loopback".

`switch.loopback.addrs` is the canonical handle every link clones (`link.rs:647`, `link.rs:666`, `port_settings.rs:376`), `Link::addrs` is an `Arc` clone of it in all four construction sites, and `AddrMap.global` is keyed on the bare `IpAddr` across both owner kinds. So the reader has to trace four call sites to learn that `Link::addrs` is not per-link at all, and `Switch::loopback` is a surprising root for global link state. "Links should clone and use this too" reads as advice where it is actually the invariant the single-ownership guarantee rests on.

Hoisting the field to `Switch::addrs` and leaving `loopback` as the module of loopback-specific helpers would make the ownership obvious and drop the `addrs` parameter from `Link::new`; the cost is giving `From<&Link> for LinkSpec` and `Link::settings` access to the `Switch` they currently don't need. If that trade isn't worth it right now, at least the struct doc ("The set of configured loopback addresses on the switch") and `set_loopback`'s ("Add a loopback IPv4 address", now both families) should say what these actually are.

dpd/src/port_settings.rs
```diff
@@ -522,121 +500,62 @@ impl PortSettingsDiff {
...
-    fn addr_del_v4(
+    fn addr_clear(
         ctx: &mut Context<'_>,
         link: &mut Link,
         rb: &mut Rollback,
-        addr: Ipv4Addr,
+        addr: IpAddr,
         tag: String,
     ) -> DpdResult<()> {
         trace!(ctx.log, "ipv4 del ({addr}: {tag})");
```
cfzimmerman, line 548: 🤖 This message was written by CLU

**Category: Nit**

Stale after the v4/v6 merge: this logs `ipv4 del` for IPv6 addresses too. Its counterpart at line 525 was updated to `ip add`.

Worth fixing rather than ignoring because this trace is the log line these wedges are actually diagnosed from — the add/del interleaving in `/tmp/<test>-dpd.stdout` is what tells you which family failed first.

dpd/src/addr.rs
```diff
@@ -0,0 +1,387 @@
...
+/// This upholds the following DPD address invariants:
+/// - Unicast addresses have exactly one owner across loopback and all links.
+/// - Such an address is in this soft state IFF it's in an asic table.
+///
+/// # Synchronization
+///
+/// Any method with a `switch` parameter will lock asic tables.
+/// Never hold an asic table lock while calling such a method.
+//
+// Design choice: A single `HashMap<IpAddr, ...>` is sufficient for
```
cfzimmerman, line 77: to do: we could also shard IPv4 and IPv6 entries

dpd/src/link.rs
cfzimmerman: Doing `let _ = port_ip::ipv4_delete_many(...)` was a bug. If asic deletion fails and we still delete the soft state, then we leak the table entry. So the new error response is intentional and a good thing imo.

However, it does seem reasonable to try each address and return the first failure if one occurred. Updated.

asic/src/chaos/mod.rs
cfzimmerman: Updated to use `self.port_to_asic_id(port_hdl)`

dpd/src/addr.rs
cfzimmerman: Updated to move the owner check before the tag check.

dpd/src/table/mac.rs
cfzimmerman: Removed the outdated comment.

dpd/src/api_server.rs
cfzimmerman: TODO: This was indeed hacky. Make a separate PR with a new API endpoint to get the plumbed state of the link. I think that's useful for debugging anyway. Can be a variant of LinkSettings.

dpd/src/table/port_ip.rs
cfzimmerman: It's a fair criticism that this single-use function has a branch that should never be taken. However, conceptually I appreciate that branch existing. I would change this if it were in the same module as `AddrMap`, but I think it's good to keep as long as they're separate.

Additionally, I'm only avoiding this on `AddrMap` `Entry::Occupied` out of distrust for the SDE. Leaving this as-is gives a one-line change if we decide to start using it on the `Occupied` branch as well.

Prefer to leave as-is.

dpd/src/addr.rs
cfzimmerman: TODO::cory can we cache the asic id?

dpd-client/tests/chaos_tests/port_settings.rs
cfzimmerman: TODO::cory: return most recent error on timeout

dpd-client/tests/chaos_tests/port_settings.rs
cfzimmerman: todo::cory: update the doc comment and include a second address in the final port_settings_apply

dpd/src/addr.rs
cfzimmerman: This advocates an even worse maintainability issue, which is allowing the invariant to silently be broken. Punishing assertions during review just means code doesn't have assertions. A bug is either loud or silent, and in CI-tested code I want the bug to be loud. I accept if a human reviewer advocates for the less-optimized single-map design to not need a new invariant, but I consider it untenable to keep an invariant like this without asserting it in the code.

dpd/src/port_settings.rs
cfzimmerman: TODO::cory: update log

